Issue #3, opened 2016-03-14
VerifyWithJWK "use" and "kty" issues


saracen:
Just wondering if this logic: https://github.com/lestrrat/go-jwx/blob/76ff5fd46c773bcf15f019da4b5247780efec796/jws/jws.go#L184 is also supposed to deal with a JWK that has the use "sig"?

Also, the specification states that the "alg" parameter of a JWK can be optional. Azure AD's OpenID implementation does not send this. Their own library seems to use RSA256 by default if the kty is RSA. Do you believe it is safe to always fallback to one of the "recommended" algorithms defined by the specification when the "alg" parameter is blank?

lestrrat:
As for the first point, good point. I don't think I was thinking about doing signature verification when I wrote this.

for the second point, perhaps it would be best to be allow the coder to specify what the default is. What is your say on this matter?

PR #4, opened 2016-04-22
accept "sig" use of JWK for verifying

i110:
As mentioned in https://github.com/lestrrat/go-jwx/issues/3, I thought `VerifyWithJWK` method should accept `sig` use value to verify JWS.
How do you think about it?

### DIFF
--- a/jws/jws.go
+++ b/jws/jws.go
@@ -181,10 +181,6 @@ func VerifyWithJWK(buf []byte, keyset *jwk.Set) ([]byte, error) {
 	}
 
 	for _, key := range keyset.Keys {
-		if u := key.Use(); u != "" && u != "enc" {
-			continue
-		}
-
 		keyval, err := key.Materialize()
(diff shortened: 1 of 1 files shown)


lestrrat:
I get #3, but IRC this line exists precisely because OpenID foundation's test suite required me to find the correct key from a list of signatures, so I don't think I want to merge it as is.

I think the correct behavior would be "look for a key that uses 'enc' if nothing is specified, but if the user specifies to use a certain key ID or type, use that instead"

lestrrat:
@i110 see if 8a5d498 does it for you?

lestrrat:
I'm going to merge this, open a new issue if you still have problems

lestrrat:
sorry, I messed up. I merged my changes, and discarded yours.

i110:
@lestrrat sorry for my late reply. Your impl is great and meets my purpose. Thanks a lot.

PR #7, opened 2016-04-27
Start converting to github.com/pkg/errors

lestrrat:


fastreles:
Any chance of landing this?

lestrrat:
@fastreles With proper prodding, yes. :)

For the record, I'm not sure this covers all of the cases, but at least I stopped returning bare errors.

PR #8, opened 2016-06-22
Fix: Serialization edge case

gcallaghan:
Setting the bytes of an ECDSA coordinate can result in a length mismatch
if the octet string is not properly padded.

### DIFF
--- a/jws/interface.go
+++ b/jws/interface.go
@@ -155,3 +155,30 @@ type EcdsaVerify struct {
 type HmacVerify struct {
 	signer *HmacSign
 }
+
+// JWKAcceptor decides which keys can be accepted
+// by functions that iterate over a JWK key set.
+type JWKAcceptor interface {
+	Accept(jwk.Key) bool
+}
+
+// JWKAcceptFunc is an implementation of JWKAcceptor
+// using a plain function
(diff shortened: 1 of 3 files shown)


willmcg:
More specifically in this case, elliptic curve public key coordinates generated by many libraries occasionally generate X or Y coordinates where:

`(key.X.BitLen() + 7) / 8 < curve.Param().BitSize / 8`

This difference is usually only by a single byte but it triggers the equal byte length validation check to fail. This change left pads the octet strings with zero bytes up to the expected octet string length for the curve.

For reference, a problematic EC P-384 public key we encountered was the following:

- Curve P-384
- X = 123520477547912006148785171019615806128401248503564636913311359802381551887648525354374204836279603443398171853465
- Y = 13515585925570416130130241699780319456178918334914981404162640338265336278264431930522217750520011829472589865088261

In this case, len(X.Bytes()) is 47 instead of the expected 48 for curve P-384.

lestrrat:
@gcallaghan @willmcg Would it be possible for you guys to include test case for this?

lestrrat:
@gcallaghan what happened?

lestrrat:
Oh, you opened a new PR. Gotcha.

PR #9, opened 2016-06-23
Fix: Serialization edge case

gcallaghan:
Setting the bytes of an ECDSA coordinate can result in a length mismatch
if the octet string is not properly padded.


lestrrat:
note: more explanation on #8

lestrrat:
@gcallaghan Thanks!

PR #12, opened 2016-07-29
Looks like a typo, "jtu" for "jti".

lpar:
According to the spec document the JWT ID is in the "jti" claim, not "jtu".


lestrrat:
Yep, absolutely right.

PR #15, opened 2016-08-20
Add key_ops support for jwk/Key (g|s)etter interface

kwry:
Value of key_ops of accessor was added because it was not implemented.
If the dare of unsupported, please close this PR.

### DIFF
--- a/jwk/header_test.go
+++ b/jwk/header_test.go
@@ -13,6 +13,7 @@ func TestHeader(t *testing.T) {
 	values := map[string]interface{}{
 		"kid":     "helloworld01",
 		"kty":     jwa.RSA,
+		"key_ops": []KeyOperation{KeyOpSign},
 		"use":     "sig",
 		"x5t":     "thumbprint",
 		"x5t#256": "thumbprint256",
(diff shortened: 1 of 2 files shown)


jwk/header_test.go
```diff
@@ -34,6 +35,11 @@ func TestHeader(t *testing.T) {
 		if !assert.Equal(t, v, got, "values match '%s'", k) {
 			return
 		}
+
+		err = h.Set(k, v)
+		if !assert.NoError(t, err, "Set works for '%s'", k) {
+			return
+		}
 	}
```
kwry: add simple setter test with getter test

jwk/header_test.go
```diff
@@ -13,6 +13,7 @@ func TestHeader(t *testing.T) {
 	values := map[string]interface{}{
 		"kid":     "helloworld01",
 		"kty":     jwa.RSA,
+		"key_ops": []KeyOperation{KeyOpSign},
 		"use":     "sig",
 		"x5t":     "thumbprint",
```
kwry: https://tools.ietf.org/html/rfc7517#section-4.3
>   The "use" and "key_ops" JWK members SHOULD NOT be used together;

It is `SHOULD NOT` , but was acceptable because it is for the test.

kwry:
Thank you for quick response!!

lestrrat:
I was so sure I sent out a "Thank you" to this thread on my way home a few hours ago, but it's no where to be found... sigh.
Anyway, thank you!

PR #16, opened 2016-09-02
Initialize the EssentialHeader of ECDSA JWKs.

willmcg:
The jwk.NewEcdsaPublicKey and jwk.NewEcdsaPrivateKey functions
are not initializing the EssentialHeader of the EcdsaPublicKey struct
which means they do not have an initialized KeyType (jwa.EC) and
attempts to set header parameters with Set() cause a panic. This
initialization is done correctly for RSA keys.

This commit fixes the initializion of the EssentialHeader for JWK
ECDSA keys and sets their KeyType parameter to jwa.EC. A unit test
is also implemented to verify correct behavior.


lestrrat:
Than you very much!

PR #17, opened 2016-09-04
Fix JWT parsing of JWT NotBefore (nbf) header parameter.

willmcg:
The Construct() method for a JWT EssentialClaim has an incorrect
error check comparison (simply a != instead of an ==) for the
GetString() of the "nbf" parameter. This causes the nbf value to
not be parsed when parsing a JWT. The subsequent check for empty
value prevented the error from being exposed. This fix corrects
the equality condition.

The Set() method for nbf has also been corrected to round the
time.Time argument to the nearest second to match the granulity
of the time format string in the JWT JSON. This rounding allows
the marshal/unmarshal roundtrip to JSON to be lossless. Without
it the unit test comparison fails because the initial time value
(from time.Now) has a nanoseconds component that gets stripped
off and lost when encoding to JSON. The time value is also converted
to UTC to match the timezone of the encoded nbf string.

The unit test was also updated to verify this fix by adding the
nbf parameter to the test ClaimSet. I also correct the nonce
parameter to be jti and added a private claim for a little more
coverage.

### DIFF
--- a/jwt/jwt.go
+++ b/jwt/jwt.go
@@ -76,7 +76,7 @@ func (c *EssentialClaims) Construct(m map[string]interface{}) error {
 	c.IssuedAt, _ = r.GetInt64("iat")
 	c.Issuer, _ = r.GetString("iss")
 	c.JwtID, _ = r.GetString("jti")
-	if v, err := r.GetString("nbf"); err != nil {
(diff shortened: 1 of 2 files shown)


lestrrat:
This is excellent, thank you!

PR #22, opened 2017-02-09
Fix bad trimming of compact JWS signatures.

willmcg:
The decoding of compact form JWS tokens incorrectly performs a
TrimRight to strip trailing zero bytes from the token signature
after decoding from base64. It should not do this.

I found that about 1% of the JWS tokens we generated using
another tool were failing signature verification by go-jwx.
These JWS tokens were valid and correctly formatted and after
some spelunking into the go-jwx code we discovered this trim
of trailing zeroes from the compact signature was the problem.
This specifically fails when the generated signature (ECDSA
P-384 in this case) actually had a trailing zero byte in the
serialized S component of the signature.

This patch removes the offending TrimRight. A unit test has
also been added to verify one of the specific instances of
a failing JWS token from my testing. This trimming does not
appear to be done for JSON encoded JWS tokens.

### DIFF
--- a/jws/jws.go
+++ b/jws/jws.go
@@ -353,7 +353,6 @@ func parseCompact(buf []byte) (*Message, error) {
 	if _, err := enc.Decode(signature, parts[2]); err != nil {
 		return nil, err
 	}
-	signature = bytes.TrimRight(signature, "\x00")
 
 	s := NewSignature()
 	s.Signature = signature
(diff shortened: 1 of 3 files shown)


lestrrat:
thanks, looking into the test failure.

lestrrat:
The monotonic clock added in the golang tip is causing problems checking for equality.
pondering how to fix...

lestrrat:
@willmcg Hi, I fixed the monotonic issue, so please rebase and let me know when you are done with this PR. Thanks!

willmcg:
Rebased and everything is looking good. Thanks.

lestrrat:
Thank you!

PR #47, opened 2018-02-26
Change jwk "headers" to "parameters"

zamicol:
As per rfc 7517 section-4


zamicol:
Github says I closed this.  Not sure how, is that because of the recent commit?

commit 9f0043693f87acb8c2508657f7fc978d0bef7165

lestrrat:
No, it was closed since the first time I saw it, which was before my commit

lestrrat:
I really don't care for the terminology, and if you want to change it that's fine, but this breaks backwards compatibility only for the sake of renaming, which is not something I can accept.

The "main" implementation can be the new one, and the docs may state so, but you need to at least make the old name work for the time being and write tests for it for this to be considered

PR #53, opened 2018-05-11
copy only lower bits in ecdsaSignFunc

### LINKED ISSUE #52: Intermittently generating an incorrect signature from jwt.SignedString
I wrote a test to reproduce: 

```
func Test_Signature(t *testing.T) {
    for i := 0; i < 1000; i++ {
        tok := jwt.New()
        priv, err := ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
        assert.FatalError(t, err)
        pub := &priv.PublicKey
        assert.FatalError(t, err)
        s, err := tok.Sign(jwa.ES256, priv)
        assert.FatalError(t, err)
        _, err = jws.Verify([]byte(s), jwa.ES256, pub)
        assert.FatalError(t, err)
    }
}
```

This breaks consistently within 1000 iterations whether I use the verify method from `jws` or one from another package.

Warashi:
This fix #52 
The problem is dst of copy is larger size than src.
This PR adjusts dst size to fit with src.

### DIFF
--- a/jws/sign/ecdsa.go
+++ b/jws/sign/ecdsa.go
@@ -41,7 +41,7 @@ func makeECDSASignFunc(hash crypto.Hash) ecdsaSignFunc {
 		rb := r.Bytes()
 		vb := v.Bytes()
 		copy(out[keysiz-len(rb):], rb)
-		copy(out[keysiz*2-len(vb):], v.Bytes())
+		copy(out[keysiz*2-len(vb):], vb)
 		return out, nil
 	})
 }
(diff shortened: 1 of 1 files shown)


dopey:
First off, awesome find! Maybe change the last `v.Bytes() -> vb`. I just ran on 1mil iterations, no errors.

Warashi:
oops, fixed `v.Bytes() -> vb` and pushed

lestrrat:
You rock! confirmed on my end too. Thank you @Warashi !

PR #58, opened 2018-09-27
Fix wrong key size of ECDH-ES + AES key wrap

yaosiang:
## Problem
When using `ECDH_ES_A256KW` to encrypt plaintext, error `invalid key size 64` will be thrown.

## Cause
In jwe.Encrypt(), content key size in switch case `jwa.ECDH_ES_A256KW`, `jwa.ECDH_ES_A192KW`, `jwa.ECDH_ES_A128KW` is wrong.

```golang
case jwa.ECDH_ES_A128KW, jwa.ECDH_ES_A192KW, jwa.ECDH_ES_A256KW:
    pubkey, ok := key.(*ecdsa.PublicKey)
    if !ok {
        return nil, errors.New("invalid key: *ecdsa.PublicKey required")
    }
    keyenc, err = NewEcdhesKeyWrapEncrypt(keyalg, pubkey)
    if err != nil {
        return nil, errors.Wrap(err, "failed to create ECDHS key wrap encrypter")
    }
    keysize = contentcrypt.KeySize() // content key size is wrong
}
```

## Solution
Key size divided by 2.

> Thanks my colleague @crownych, he found this bug.

### DIFF
--- a/jwe/jwe.go
+++ b/jwe/jwe.go
@@ -68,7 +68,7 @@ func Encrypt(payload []byte, keyalg jwa.KeyEncryptionAlgorithm, key interface{},
 		if err != nil {
 			return nil, errors.Wrap(err, "failed to create ECDHS key wrap encrypter")
 		}
-		keysize = contentcrypt.KeySize()
+		keysize = contentcrypt.KeySize() / 2
 	case jwa.ECDH_ES:
 		fallthrough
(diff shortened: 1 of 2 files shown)


lestrrat:
wow, oops, I was still testing it out. I'm going to revert it once just to make sure I actually looked at it, and then will proceed with merging.

lestrrat:
Checked it out. Thank you!

yaosiang:
Very quickly !

PR #64, opened 2019-03-22
jws.Sign() removes whitespace and newlines from JSON

### LINKED ISSUE #62: jws.Sign() removes whitespace and newlines from JSON
I would like to use jws.Sign() to parse any JSON payloads such as https://tools.ietf.org/html/rfc7515 Example A.1, A.2, etc 

Unfortunately that is not possible because jws.Sign() strips whitespace, newlines and carriage return from the JWS Protected Header.  Basically, jws.Sign() removes all whitespace from the  JSON. So, instead of signing:

```
     {"typ":"JWT",
      "alg":"HS256"}

[123, 34, 116, 121, 112, 34, 58, 34, 74, 87, 84, 34, 44, 13, 10, 32,
   34, 97, 108, 103, 34, 58, 34, 72, 83, 50, 53, 54, 34, 125]
```

It is signing (no whitespace, no new lines):

```
{"typ":"JWT","alg":"HS256"}
```

jws.Sign() should be able to take the protected headers "as is", without trying to add to a structure and then serialize it. If it changes the JWS headers as presented, it will generate different signatures from other systems or that the original document holder expected.

repenno:
fixes #62
Added idea folders to gitignore

This PR adds a new function that serializes protected headers "as is", without formatting  or removing whitespace.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,8 @@ _testmain.go
 *.exe
 *.test
 *.prof
+
+# IDE
+.idea
+.vscode
+.DS_Store
--- a/buffer/buffer.go
+++ b/buffer/buffer.go
@@ -1,5 +1,5 @@
 // Package buffer provides a very thin wrapper around []byte buffer called
-// `Buffer`, to provide functionalitites that are often used wthin the jwx
+// `Buffer`, to provide functionalities that are often used within the jwx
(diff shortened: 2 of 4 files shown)


lestrrat:
Thank you very much!

PR #73, opened 2019-04-04
Helper function to avoid casts to jws.Verify()

repenno:


lestrrat:
LGTM. added comments, and merged

Issue #79, opened 2019-04-06
Low test coverage


repenno:
I used the following to examine coverage.

go test -v ./... -coverprofile=coverage.out
go tool cover -html=coverage.out

- Many components are below 70% and quite a few under 50%.  
- Some functions do not seem to be used throughout the code and have no coverage.  In my copy of the repo I just removed most of these functions, rerun the tests and they all passed.

lestrrat:
Which functions are we talking about that are not used? I just reran [unused](https://github.com/dominikh/go-tools) and it didn't report anything :/ I'm sure the tool isn't picking up everything, but if you have a list already, please let me know

repenno:
If you look at the coverage reported by the go tool, there are quite a few. I am not sure how "unused" work, but being exported or implementing an interface should not count if the function is not actually called at all. 

How about in:

1 - symmetric.go

(k *SymmetricKey) UnmarshalJSON(data []byte) 

2 - option.go

func WithPretty(b bool) Option {
        return option.New(optkeyPrettyJSONFormat, b)
}

func WithHeaders(h Headers) Option {
        return option.New(optkeyHeaders, h)
}

3 - date.go

func (n *NumericDate) UnmarshalJSON(data []byte) error

These functions show zero coverage and I can not see usages for them.

lestrrat:
Interesting, I'd thought `unused` would report something like `WithPretty`, but I guess not.
Thanks. will look into it.

repenno:
"  - (8.3) All interface methods are marked as used, even if they never get
    called. This is to accomodate sum types (unexported interface
    method that must exist but never gets called.)"

Looking at other explanations it seems they mark a lot of stuff as used even though it is never actually called at all.

lestrrat:
After fixing a couple of those UnmarshalJSON things in `jwt`, I think (at least in `jwt`) what happened was:

1. Initially everything was done via `UnmarshalJSON`
2. Then I realized that the tokens can be flattened and normalized in various patterns
3. ... and then I moved to Unmarshal'ing everything to a map/slice/string first, and then used heuristics to deduce the actual types to convert, and then used `Accept` interface to receive the values

So yeah, the story might be different in other packages, but I'm guessing most of those UnmarshalJSON's can go away now.

lestrrat:
Okay, so at least I went ahead and removed some of the obvious choices including the ones you mentioned.
I'm going to stop here for the day.

[...]

PR #83, opened 2019-04-07
Remove unused options

lestrrat:
git blame / git log tells us that they had their use way back when, but not anymore

refs #79


lestrrat:
note: `WithPretty` was a legacy artifact. `WithHeaders` may be untested, but is still referenced.

PR #99, opened 2019-04-10
coverage for jwk package

lestrrat:


lestrrat:
Now, I'm sure I was doing some more extra work on my local machine, but alas my machine has been taken for repair, and I don't have commits elsewhere. I guess I'm just going to merge as is

PR #100, opened 2019-04-10
Low coverage and unused functions in jws

### LINKED ISSUE #89: Low coverage and unused functions in jws/headers.go
Function JWKSetURL() is not used or covered

https://github.com/lestrrat-go/jwx/blob/9871a5674c2ca011d6156f4cd5200301be523104/jws/headers.go#L78

None of these functions are used or covered

https://github.com/lestrrat-go/jwx/blob/9871a5674c2ca011d6156f4cd5200301be523104/jws/headers.go#L103-L122

None of these switch cases are covered

https://github.com/lestrrat-go/jwx/blob/9871a5674c2ca011d6156f4cd5200301be523104/jws/headers.go#L174-L196

None of these switch cases are covered

https://github.com/lestrrat-go/jwx/blob/9871a5674c2ca011d6156f4cd5200301be523104/jws/headers.go#L249-L272

None of these lines are covered

https://github.com/lestrrat-go/jwx/blob/9871a5674c2ca011d6156f4cd5200301be523104/jws/headers.go#L379-L395

https://github.com/lestrrat-go/jwx/blob/9871a5674c2ca011d6156f4cd5200301be523104/jws/headers.go#L343-L359

repenno:
fixes #89
- During fixing noticed that marshal and unmarshal were not reversible,
  i.e., if you marshaled with existing function, unmarshal would fail.
  This is due to the fact that []string become []interface{} and those
  are not the same:
  https://stackoverflow.com/questions/32530952/should-i-be-able-to-type-assert-a-slice-of-string-maps
  https://golang.org/doc/faq#convert_slice_of_interface
- Simplified significantly (un)marshal by using JSON tags. This
  increased coverage and made code simpler
- Removed unused code
- Simplified interface by removing unused or duplicate functions
- Increase coverage of verify and sign functions to 80%+
- Increased test coverage of headers.go to 80%+
- Overall coverage (if we removed pdebug) would 90%+

thanks,

### DIFF
--- a/cmd/jwx/jwx.go
+++ b/cmd/jwx/jwx.go
@@ -142,7 +142,8 @@ func doJWK() int {
 			log.Printf("%s", l)
 		}
 
-		if _, err := jws.Verify(buf.Bytes(), sig.ProtectedHeaders().Algorithm(), pubkey); err == nil {
+		alg := sig.ProtectedHeaders().Algorithm()
+		if _, err := jws.Verify(buf.Bytes(), alg, pubkey); err == nil {
 			log.Printf("=== Verified with signature %d! ===", i)
 		}
 	}
--- a/jws/headers.go
+++ b/jws/headers.go
@@ -2,10 +2,10 @@
 package jws
 
 import (
-	"encoding/json"
 	"github.com/lestrrat-go/jwx/jwa"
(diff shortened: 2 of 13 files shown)


lestrrat:
`reflect` is unnecessary in headers.go, so I'll merge after I fix that bit

PR #103, opened 2019-04-15
Add go.mod

lestrrat:


go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/lestrrat-go/jwx
+
+go 1.12
+
+require (
+	github.com/lestrrat-go/pdebug v0.0.0-20180220043849-39f9a71bcabe
```
trivigy: @repenno Same semantic versioning applies here. Do you see how this library depends on specific commits of other sub `github.com/lestrrat-go` projects?! This will completely break if someone decides to one day `push --force` or squash commits on master. I have seen that happen with a few libraries out there in the wild. SUPER annoying.

repenno: In general I agree that dependencies (specially transitive dependencies) are no fun in Go and we need semver. 

More specifically, I would completely remove pdebug dependency and opened an issue for that.  I did on my copy of the repo and it reduced code, reduced complexity and upped coverage.

trivigy: That is awesome. I totally agree with everything you said. Any chance you could PR the `pdebug` removed code?

repenno: https://github.com/lestrrat-go/jwx/pull/111

PR #105, opened 2019-04-15
Low coverage and unused functions in token_gen.go

### LINKED ISSUE #85: Low coverage in token_gen.go
https://github.com/lestrrat-go/jwx/blob/9871a5674c2ca011d6156f4cd5200301be523104/jwt/token_gen.go

Get() has very low coverage
functions JwtID(), NotBefore() and Expiration() are not used
Error pats in Set() are not covered.

repenno:
fixes #85

1 - Overall Changes

- Removed Token Top Level (Un)Marshal and added JSON Tags for everything
  - This simplified code and increased coverage
- Added specific (un)Marshal for NumericDate so that we always return a
  number according to the definition of NumericDate on
  https://tools.ietf.org/html/rfc7519#page-6

2 - Token Behavior

- We should only accept audience(aud) claims as an array, even if it
  is an array of one element. This greatly simplifies code and coverage

3 - Bugs

- If we add private claims, we need to indicate as such when getting
  tokens. See example_test.go. We get this for free by using JSON tags
  for (un)Marshal.

- When creating a JWT the New() function should behave just
  like "var token Token". the New() function was creating an empty
  private PrivateClaims map and that would break JSON (un)Marshal. The
  PrivateClaims should be created on-demand much like existing JWS
  headers code.

4 - Testing and Coverage

- Added string_test.go to properly test stringlist implementation
- Moved date testing to date_test.go so we get proper coverage
- Moved token specific tests from jwt_test.go to token_gen_test.go
- Moved verify specific tests from jwt_test.go to verify_test.go

Coverage is at 90%+ for Token

thanks,

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,5 @@ _testmain.go
 .idea
 .vscode
 .DS_Store
+
+coverage.out
--- a/jwt/date.go
+++ b/jwt/date.go
@@ -2,6 +2,7 @@ package jwt
 
 import (
 	"encoding/json"
+	"strconv"
 	"time"
 
 	"github.com/pkg/errors"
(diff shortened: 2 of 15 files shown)


lestrrat:
sorry for the wait! thanks!

PR #111, opened 2019-04-19
Can we consider removing pdebug dependency?

### LINKED ISSUE #107: Can we consider removing pdebug dependency?
I use an IDE and can step and debug the entire code without pdebug. I removed the entire pdebug dependency in a branch to experiment and coverage overall went up 12%  and the number of lines of code went down considerably.

repenno:
fixes #107


lestrrat:
yep. I really needed it while I was figuring it out, but I guess it can go

PR #112, opened 2019-04-23
Support single audience json field

atotto:
We should consider that parse single audience json field for other jwt library.

breaking change: #105

Thanks.

### DIFF
--- a/jwt/example_test.go
+++ b/jwt/example_test.go
@@ -47,9 +47,7 @@ func ExampleSignAndParse() {
 	}
 	// OUTPUT:
 	// {
-	//   "PrivateClaims": {
-	//     "foo": "bar"
-	//   }
+	//   "foo": "bar"
 	// }
 }
 
@@ -80,10 +78,8 @@ func ExampleToken() {
 	//     "Golang Users"
 	//   ],
 	//   "iat": 233431200,
-	//   "sub": "https://github.com/lestrrat-go/jwx/jwt",
-	//   "PrivateClaims": {
-	//     "privateClaimKey": "Hello, World!"
-	//   }
+	//   "privateClaimKey": "Hello, World!",
+	//   "sub": "https://github.com/lestrrat-go/jwx/jwt"
 	// }
 	// aud -> '[Golang Users]'
 	// iat -> '1977-05-25T18:00:00Z'
--- a/jwt/interface.go
+++ b/jwt/interface.go
@@ -1,6 +1,8 @@
 package jwt
 
 import (
+	"encoding/json"
+	"reflect"
 	"time"
 )
 
@@ -9,4 +11,28 @@ type StringList []string
 // NumericDate represents the date format used in the 'nbf' claim
 type NumericDate struct {
 	time.Time
+}
+
+func (l *StringList) UnmarshalJSON(b []byte) error {
+	var v interface{}
+	if err := json.Unmarshal(b, &v); err != nil {
+		return err
+	}
+	switch value := v.(type) {
+	case []interface{}:
+		list := make([]string,0,len(value))
+		for _, v := range value {
+			element, ok := v.(string)
+			if ok {
+				list = append(list, element)
+			}
+		}
+		*l = list
+	case string:
+		*l = []string{value}
+	default:
+		return &json.InvalidUnmarshalError{reflect.TypeOf(v)}
+	}
+
+	return nil
 }
--- a/jwt/interface.go
+++ b/jwt/interface.go
@@ -20,7 +20,7 @@ func (l *StringList) UnmarshalJSON(b []byte) error {
 	}
 	switch value := v.(type) {
 	case []interface{}:
-		list := make([]string,0,len(value))
+		list := make([]string, 0, len(value))
 		for _, v := range value {
 			element, ok := v.(string)
 			if ok {
--- a/jwt/internal/cmd/gentoken/main.go
+++ b/jwt/internal/cmd/gentoken/main.go
@@ -160,11 +160,9 @@ func generateToken() error {
 	fmt.Fprintf(&buf, "\n// by embedding the jwt.Token type in it")
 	fmt.Fprintf(&buf, "\ntype Token struct {")
 	for _, field := range fields {
-		jsonTag := "`" + `json:"` + field.JSONKey + `,omitempty"` + "`"
-		fmt.Fprintf(&buf, "\n%s %s %s // %s", field.Name, field.Type, jsonTag, field.Comment)
+		fmt.Fprintf(&buf, "\n%s %s // %s", field.Name, field.Type, field.Comment)
 	}
-	jsonTag := "`" + `json:"` + `,omitempty"` + "`"
-	fmt.Fprintf(&buf, "\nPrivateClaims map[string]interface{} %s", jsonTag)
+	fmt.Fprintf(&buf, "\nprivateClaims map[string]interface{}")
 	fmt.Fprintf(&buf, "\n}") // end type Token
 
 	fmt.Fprintf(&buf, "\n\nfunc (t *Token) Get(s string) (interface{}, bool) {")
@@ -201,9 +199,9 @@ func generateToken() error {
 		}
 	}
 	fmt.Fprintf(&buf, "\n}") // end switch
-	fmt.Fprintf(&buf, "\nif v, ok := t.PrivateClaims[s]; ok {")
+	fmt.Fprintf(&buf, "\nif v, ok := t.privateClaims[s]; ok {")
 	fmt.Fprintf(&buf, "\nreturn v, true")
-	fmt.Fprintf(&buf, "\n}") // end if v, ok := t.PrivateClaims[s]
+	fmt.Fprintf(&buf, "\n}") // end if v, ok := t.privateClaims[s]
 	fmt.Fprintf(&buf, "\nreturn nil, false")
 	fmt.Fprintf(&buf, "\n}") // end of Get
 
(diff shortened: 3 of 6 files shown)


atotto:
Fix private claim field.

repenno:
I disagree with this change, we should not go back to custom (un)marshal. You can support a single audience field by having an array of one.

atotto:
We are using the firebase authentication that has single string audience and flat custom fields.

https://tools.ietf.org/html/rfc7519#section-4.1.3 say we can use single string audience field.

repenno:
Hello Ato, 

it is a MAY requirement, meaning the weakest of all IETF. But if you have a running environment I do not want to create a problem. At the same time, putting back custom marshal is not a good solution. We should probably use json.rawmessage or custom type such as stringlist and perform a two step unmarshal of *only* the aud field and not the entire JSON.

lestrrat:
Hmm, previously I completely neglected the fact that aud could be plural, so making its type []string was definitely a good thing. But yeah, JSON based formats always have this dilemma where it allows for arbitrary containers, which is always a problem.  I'll play with it and see what feels right on my end.

lestrrat:
@repenno @atotto does this work? 99f6731

atotto:
LGTM for the audience field.

lestrrat:
@atotto I guess custom claims inadvertently got wiped out too, eh?

atotto:
@lestrrat The custom claims (private claims) deleted by #105.
So I rewrite json marshal/unmarshal for Token struct.

lestrrat:
let me ponder about that later in a separate PR

repenno:
Folks,

Custom claims are there, but you need to pass inside a special object called Private Claims. I feel this is the proper solution *unless* we implement all public and registered claims today (and forever). Otherwise we do not know what is truly private claim and what is a public claim we have not implemented.

Example: Are "name" and  "phone_number" private claims? No, we have not just implemented them, but they are part of IANA:  https://www.iana.org/assignments/jwt/jwt.xhtml#claims. We need a way to deterministically determine if something is a private claim or not.

   "A producer and consumer of a JWT MAY agree to use Claim Names that
   are Private Names: names that are not Registered Claim Names
   (Section 4.1) or Public Claim Names (Section 4.2).  Unlike Public
   Claim Names, Private Claim Names are subject to collision and should
   be used with caution."

lestrrat:
Well, but the usefulness of a token will severely be limited unless one can use possibly un-implemented token keys seamlessly. IIRC tokens generate by Google, for example, use private claims that users actually kinda need to use.

I think my mistake was that I allowed the API to access pre-defined fields as `t.ExportedFieldName` AND `t.Get(fieldNameKey)`. I should've just made it

```
// for all "well known" fields:
v = t.Audience()
v = t.IssuedAt() 
// etc...

// for everything else (but including the well known fields)
v, ok := t.Get(jwt.AudienceKey)
list := v.([]string)

v, ok := t.Get("mySuperPrivateClaim")
s := v.(string) // or whatever

// and json.Marshal(t) -> { "aud": ... , "iat": ..., "privateClaim": ... }
```

repenno:
I also want to support PrivateClaims, so we agree. I believe we are discussing how to better support it. I would suggest part of this effort then:  

1 - add all existing IANA claims to the data structure. It really does not matter if we do anything with them, but we JSON tag for accurate and easily (un)marshal. We will be able to determine what is private with more accuracy, still not perfect of course, but much better.
2 -We need to make sure claims do not collide as suggested in the RFC. The JSON we receive might not have been sanitized or might have been composed from different sources. Private claims can collide amongst themselves as well.

lestrrat:
Agreed. Prod me again if you guys haven't heard from me in 24 hrs

lestrrat:
re collision:

>2 -We need to make sure claims do not collide as suggested in the RFC. The JSON we receive might not have been sanitized or might have been composed from different sources. Private claims can collide amongst themselves as well.

Currently I don't see any way to really detect collisions in a given JSON object from Go.
for single values, the last one wins, and values with nested maps, values will be merged.

To avoid this we would need a custom JSON parser. I dunno it that's worth it.

repenno:
If anything I am trying to get us away from custom parsers unless we really need to (re: aud). But I am surprised that is the implementation underneath. I would think a JSON with duplicate keys would generate an error. I guess we need to live with this caveat.

lestrrat:
for your reference: https://github.com/golang/go/issues/24415 :)

lestrrat:
After tinkering around a bit, I think the sane way to incorporate all IANA claims is to divide them up in to logical chunks and do something like

```
package openid
const NameKey = "name"
func Name(t *jwt.Token) string {
   if v, ok := t.Get(NameKey); ok { 
      return v.(string)
   }
   return ""
}
```

```
import (
  "github.com/lestrrat-go/jwx/jwt"
  "github.com/lestrrat-go/jwx/jwt/openid"
)

var t jwt.Token
json.Unmarshal(data, &t)
name := openid.Name(t)

openid.Set(t, openid.NameKey, "new name")
```

repenno:
golang/go#24415 . It is weirder than I thought!

Dividing into logical chunks is ok. I am assuming you will create a struct where you embed all other structs into it so we can easily (un)marshal, correct? Something like:

`
struct Token {
OpenIdClaims
Rfc7519Claims
NotSureWhatToCallTheOtherClaims
...
}
`

lestrrat:
I kind of want to NOT have to embed everything, as I think that the majority of the cases does not require the use of all claims. My intention with that API is to give the user a type-safe way to operate on known types ONLY when necessary, and allow the user to work with only the minimal version most of the time

repenno:
Makes sense, but how you propose to (un)marshal claims that might have a
mix of openid, rfc7519, rfc8055, amongst others? Example, a JSON like

{
  "name": "my full name",
  "iss": "http://www.example.com",
  "sip_from_tag": "some_caller@someplace.com",
  "private_claims": "..."
}










thanks,

lestrrat:
sorry, I pushed the comment button in haste, and am writing a longer version now. please hold

lestrrat:
> Dividing into logical chunks is ok. I am assuming you will create a struct where you embed all other structs into it so we can easily (un)marshal, correct? Something like:
>
>struct Token {
>OpenIdClaims
>Rfc7519Claims
>NotSureWhatToCallTheOtherClaims
>...
>}

No, that is not the intention. 

Let me layout the implementation that I'm thinking: The token will look like what it is as of c9557ef, which is 

```
type Token struct {
  audience      StringList             `json:"aud,omitempty"` // https://tools.ietf.org/html/rfc7519#section-4.1.3
  expiration    *NumericDate           `json:"exp,omitempty"` // https://tools.ietf.org/html/rfc7519#section-4.1.4
  issuedAt      *NumericDate           `json:"iat,omitempty"` // https://tools.ietf.org/html/rfc7519#section-4.1.6
  issuer        *string                `json:"iss,omitempty"` // https://tools.ietf.org/html/rfc7519#section-4.1.1
  jwtID         *string                `json:"jti,omitempty"` // https://tools.ietf.org/html/rfc7519#section-4.1.7    notBefore     *NumericDate           `json:"nbf,omitempty"` // https://tools.ietf.org/html/rfc7519#section-4.1.5
  subject       *string                `json:"sub,omitempty"` // https://tools.ietf.org/html/rfc7519#section-4.1.2
  privateClaims map[string]interface{} `json:"-"`
}
```

The only reason that `Token` is not just a `map[string]interface{}` is because the 7 claims originally described in RFC7519 are so frequently used that it's kind of silly to have to type-check them every time we use them. Otherwise we could've just shoved everything in a `map[string]interface{}`

Now, because the JWT may (and most likely will) contain more fields than those 7, we have the `privateClaims` which slurps the remaining fields not covered. These we don't necessarily know the type of, and so we shove them into a `map[string]interface{}`

Looking at https://www.iana.org/assignments/jwt/jwt.xhtml#claims, there are certainly fields that seem to be widely in use, but there are also others that one may almost never have to interact with depending on usage. I do want to avoid cluttering up the main API just to be complete.

One way to separate the main `jwt.Token` from less-frequently used API would be to separate out the package, and possibly create a `openid.Token`, but that doesn't solve the problem when one wants to use every possible claim listed in the above IANA site, as it's not going to be possible to mix every package.

My proposal is then to just accept to have everything stored in the original `jwt.Token` above. Most of the fields in the IANA list will be stored under `privateClaims`, but we give users a type-safe API to both set and get each field (i.e. for `name` claim, `func Name(t *jwt.Token) string` to retrieve, and `openid.Set(t *jwt.Token, key string, value interface{}) error` to set).

This will allow us to not have to modify any of the `jwt.Token` API, while giving users a simpl-er way to access these extra registered claims.

WRT marshal/unmarshal, you shouldn't have to worry, as contents in `privateClaims` will automatically be `json.Marshal`ed, and merged with the rest of the `jwt.Token` JSON value.

repenno:
okay, but I am back to my original question. What happens if we get a JSON
like:

{
  "name": "my full name",
  "iss": "http://www.example.com",
  "sip_from_tag": "some_caller@someplace.com",
  "private_claims": "..."
}

"name" is not private, it is private for our implementation only. Therefore
for somebody to pass that field under the PrivateClaims object we need to
document something like "any claim outside rfc7519 needs to be passed
inside the private claims object.". Then we should be good.

I looked at other Golang libraries and they are "behind" us, we are trying
to solve an issue they have not tackled. Very interesting!

lestrrat:
No, please read the code in c9557ef. The field `private_claims` does not exist. It's flattened and merged with the main JSON object.

repenno:
okay, so how would you move fields like "name" and "sip_from_tag" into the
map? If you just perform a json.unmarshal(json, &vatToken), unknown fields
will be ignored. So, you will construct a custom (un)marhsaller?

lestrrat:
Yes, it was that way (and yeah, sorry for not checking deeper, I didn't realize you removed *that* part from my original code) and it is that way again as of c9557ef. I know you don't like it, but to support private claims properly, it must be that way.

roitk:
Hi, I'm relatively new to the library. I was wondering if it was possible to do the reverse of this? I.e. creating a new JWT with a single string audience field.

It looks like with 

 https://github.com/lestrrat-go/jwx/blob/9143661f037e46a9d2b5861fc19c4b051966b00b/jwt/token_gen.go#L148-L153

and

https://github.com/lestrrat-go/jwx/blob/9143661f037e46a9d2b5861fc19c4b051966b00b/jwt/internal/types/string.go#L15-L18

, using `Set` forces the token to be an array of strings.

PR #116, opened 2019-04-25
Reinstate private claims

lestrrat:
refs #112


lestrrat:
Regardless of how we're going to tackle things like #117, private claims should work, so merging

PR #123, opened 2019-04-28
Clean and harmonize jws/main.go/headers.go

### LINKED ISSUE #122: Clean and harmonize jws/main.go/headers.go
Remove package name from const names
Use "name" instead of "method". We do not use "method" per se anymore after all changes, i.e., there is no direct call to any method.
Use GetAlgorithm() instead of Algorithm()
Added PrivateParams as a JSON tag. But if we want flattened we can change asap.

I am trying to get all generate code to look similar and follow a pattern.

repenno:
Fixes #122

### DIFF
--- a/cmd/jwx/jwx.go
+++ b/cmd/jwx/jwx.go
@@ -142,7 +142,7 @@ func doJWK() int {
 			log.Printf("%s", l)
 		}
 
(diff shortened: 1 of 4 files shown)


repenno:
I will keep this change only to my repo.

PR #125, opened 2019-05-20
Fix signature verification

### LINKED ISSUE #61: Incorrect failed signature verification on some tokens
I just hit a situation where one user's tokens would decode and verify, and another user's would not, giving "failed to verify jws signature: failed to verify message: crypto/rsa: verification error".

I can supply a self-contained test case which compares behavior between this library and https://github.com/dgrijalva/jwt-go (which verifies both correctly) if you want to investigate.

segfault16:
jws.SplitCompact was not working for JWTs > 4kB.

This PR fixes #61.

### DIFF
--- a/jws/jws.go
+++ b/jws/jws.go
@@ -470,33 +470,39 @@ func SplitCompact(rdr io.Reader) ([]byte, []byte, []byte, error) {
 	var protected []byte
 	var payload []byte
 	var signature []byte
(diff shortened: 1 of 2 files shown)


lestrrat:
thanks!

segfault16:
You're welcome @lestrrat. That was fast ;)

PR #133, opened 2020-02-09
Added method GetAllClaims to return all the claims from a token object

ShramikGajlekar:


lestrrat:
First and foremost, thank you. However, I can't accept this because token_gen.go is a generated file, and direct edits to it will always be overwritten. if you need to edit any of the `_gen.go` files you need to find where they are generated from: in this case `jwt/internal/cmd/gentoken/main.go`.
 
wrt to the change to the API itself, I think I'd prefer something like

```
func (t *token) WalkClaims(f func(key string, value interface{}) bool) {
    for .... <snip> ... {
        value, ok := t.Get(key)
        if ok {
             if ! f(key, value) {
                 return // bail out
             }
        }
    }
}
```

or perhaps an iterator

```
type ClaimPair struct {
   Key string
   Value interface{}
}
func (t *Token) Iterator() <-chan ClaimPair {
   // return a channel that sends a ClaimPair struct for each entry
}

for claim := range t.Iterator() {
   doSomething( claim.Key,  claim.Value )
}
```

ShramikGajlekar:
Hello @lestrrat I've made changes in the jwt/internal/gentoken/main.go file. I do see the changes that you wanted me to add for extracting all the claims from the token.With all due respect I believe the approach I have taken is bit simpler and straightforward. But if you still deem the changes to api as requested by you necessary, I shall update this PR with the same.  Thank you.

lestrrat:
I already stated I don't agree.

I assume you want to deal with maps, and that's fine, but as somebody who has been writing library APIs for well over two decades, I know that somebody definitely WILL come up with a case where efficiency will matter, and at that point an API that allocates a map will feel clunky. Then I will have to add a more generic API, which will be more annoying.

PR #135, opened 2020-02-18
Implement iterator, walk, and map generating methods

### LINKED ISSUE #132: No method to return all claims.
For refreshing token, we use all the claims from previous tokens and modify the `iat` and `exp` claims. If we have several claims, it's painful to extract all claims from the token and sign a new one.

lestrrat:
fixes #132 
closes #133


lestrrat:
Alright, I botched this PR, thinking I was working on a different issue. Dang it.

PR #141, opened 2020-03-09
using payload and err variables declared in function signature

### LINKED ISSUE #140: jws.VerifyWithJKU Passes on invalid signature
I was using v0.9.0,  `jws.VerifyWithJKU()` validation worked fine. But once I updated to the newer version v0.9.1, it no longer validates the signature, returns no error. This is a serious issue. Could I get any help regarding this? 

Thank you.

leandrorondon:
fix #140

### DIFF
--- a/jws/jws.go
+++ b/jws/jws.go
@@ -363,7 +363,7 @@ func VerifyWithJWKSet(buf []byte, keyset *jwk.Set, keyaccept JWKAcceptFunc) (pay
 			continue
 		}
 
-		payload, err := VerifyWithJWK(buf, key)
+		payload, err = VerifyWithJWK(buf, key)
 		if err == nil {
 			return payload, nil
 		}


jws/jws.go
```diff
@@ -363,7 +363,7 @@ func VerifyWithJWKSet(buf []byte, keyset *jwk.Set, keyaccept JWKAcceptFunc) (pay
 			continue
 		}
 
-		payload, err := VerifyWithJWK(buf, key)
+		payload, err = VerifyWithJWK(buf, key)
```
leighmcculloch, line 366: It's interesting that the tests pass with and without this change. Could you add a test as part of this fix?

leandrorondon:
> It's interesting that the tests pass with and without this change. Could you add a test as part of this fix?

> This fix looks good to me. This also seems like a reasonably serious issue that the error is not being propagated. Is it possible to include this in a release?

@leighmcculloch Thanks for the feedback.

It's the first time I contribute to an open-source project :)
Can I make the requested changes in this same PR?

leighmcculloch:
@leandrorondon I'm not a contributor to this project so I can't speak for @lestrrat. I would think adding some tests would be good.

lestrrat:
Sorry for the delay here. Yeah, I think a test would be nice.

lestrrat:
Okay, so I finally got around to looking at the problem (sorry for warnocking #140). I don't think this PR fixes anything, and #140 certainly doesn't have anything to do with recent changes to this particular piece of code.

As for this particular piece of code, I can certainly tidy it up and I will do so

However, I don't think this fixes #140.

lestrrat:
oh wait, I'm sorry, I spoke too soon. Sorry @leandrorondon . Will update soon....

lestrrat:
I think #145 fixes is more "properly"

lestrrat:
closed by #145

PR #143, opened 2020-03-29
Add WithClaimValue to JWT Verify options

sonatard:
WithClaimValue allows you to verify any claim.

before

```go
	if err := token.Verify(
		jwt.WithClock(clock),
		jwt.WithIssuer("https://accounts.google.com"),
		jwt.WithAudience(audience),
	); err != nil {
		return xerrors.Errorf(": %w", err)
	}

	if emailVerified, ok := token.Get("email_verified"); !ok || emailVerified != true {
		return xerrors.Errorf("token not valid: email is not verified")
	}

	if email, ok := token.Get("email"); !ok || email != "email@example.com" {
		return xerrors.Errorf("token not valid: no permission %v", email)
	}
```

after
```go
	if err := token.Verify(
		jwt.WithClock(clock),
		jwt.WithIssuer("https://accounts.google.com"),
		jwt.WithAudience(audience),
		jwt.WithClaimValue("email_verified", true),
		jwt.WithClaimValue("email", "email@example.com"),
	); err != nil {
		return xerrors.Errorf(": %w", err)
	}
```

### DIFF
--- a/jwt/verify.go
+++ b/jwt/verify.go
@@ -2,6 +2,7 @@ package jwt
 
 import (
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/lestrrat-go/jwx/internal/option"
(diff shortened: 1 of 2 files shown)


lestrrat:
LGTM. Thanks!

PR #144, opened 2020-03-30
Pass context to fetch functions

sonatard:
We can use `http.NewRequestWithContext` but it can use from Go 1.13.
I don't use it this PR.

```go
	req, err := http.NewRequest(http.MethodGet, jwkurl, nil)
	if err != nil {
		return nil, errors.Wrap(err, "failed to new request to remote JWK")
	}
	req.WithContext(ctx)
```

from Go 1.13
```go
	req, err := http.NewRequestWithContext(http.MethodGet, jwkurl, nil)
	if err != nil {
		return nil, errors.Wrap(err, "failed to new request to remote JWK")
	}
```

`http.NewRequestWithContext` has performace advantage. `req.WithContext(ctx)`copy http.Request.

```go
func (r *Request) WithContext(ctx context.Context) *Request {
	if ctx == nil {
		panic("nil context")
	}
	r2 := new(Request)
	*r2 = *r //--------------------------------- copy http.Request  -------------------------------
	r2.ctx = ctx
	r2.URL = cloneURL(r.URL) // legacy behavior; TODO: try to remove. Issue 23544
	return r2
}
```

Please give me advice use http.NewRequestWithContext or not.

### DIFF
--- a/jwk/jwk.go
+++ b/jwk/jwk.go
@@ -90,7 +90,7 @@ func Fetch(urlstring string, options ...Option) (*Set, error) {
 	return nil, errors.Errorf(`invalid url scheme %s`, u.Scheme)
 }
 
-// FetchHTTP fetches the remote JWK and parses its contents
+// FetchHTTP wraps FetchHTTPWithContext using the background context.
 func FetchHTTP(jwkurl string, options ...Option) (*Set, error) {
 	return FetchHTTPWithContext(context.Background(), jwkurl, options...)
 }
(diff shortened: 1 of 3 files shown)


lestrrat:
I think the simplest way to go forward is to just use build tags, and include the new method(s) only for go > 1.13

sonatard:
Thank you for the advice!

I think If jwt provides backward compatibility for < Go 1.13, had better use req.WithContext(ctx).

- It was common before Go 1.13.
- Performance issue is not big disadvantage
- Enable to use context for cancelling and tracing before Go 1.13
- common source code is easy to maintenance

lestrrat:
I'm finally sitting down in front of this and took a long look, and I think this is fine as it is.

lestrrat:
thanks!

PR #146, opened 2020-04-10
Fixed data type of JWS["jwk"] to be JWK instead of JWKSet

reinkrul:
The datatype of the "jwk" field of JWS' currently incorrectly specified as jwk.Set. According to the spec this should be a plain JWK, rather than a set (https://tools.ietf.org/html/rfc7515#section-4.1.3).

This bug breaks compatibility with other JWS libraries.


lestrrat:
hmm, I forget why I did it this way.

lestrrat:
anyway, reading the RFC it looks like this is correct, lgtm. Thanks!

lestrrat:
Okay, for whatever reason Github did not show me that the tests were failing, and I foolishly merged the changes. Reverting.

PR #147, opened 2020-04-10
Go 1.12 modules support

reinkrul:
Add go.mod and go.sum for Go modules support. Also makes it easier to fork the repository.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,8 @@
+module github.com/lestrrat-go/jwx
+
+go 1.13
+
+require (
+	github.com/pkg/errors v0.9.1
+	github.com/stretchr/testify v1.5.1
+)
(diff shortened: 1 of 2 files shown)


lestrrat:
I know it's been a while since I took a long hard look at this, but I'm not a fan of placing go.sum in library projects. There were weird cases where it just didn't work (I'm perfectly fine with go.mod)

I wonder if this has changed?

reinkrul:
I had to lookup what go.sum actually does; it contains hashes of the specified dependencies for added safety. According to the Go docs it should be checked in (https://blog.golang.org/using-go-modules). The only problems I had with it was when merge conflicts occur; they can be tricky to resolve/

Would you rather leave it out? I tried building after deleting go.sum, and it worked fine. The file was simply recreated (I guess using the hashes of the dependencies I have locally). If decided that it shouldn't be checked in, we should add it to .gitignore.

lestrrat:
I generally like putting go.sum on applications because for an application, the build should be 100% reproducible. But I find libraries to be in a slightly different situation, because it's ultimately going to be used by a larger application, where the author of said application needs the final control.

Now, of course if everything Just Works, it's fine, but I've encountered weeeeeird cases where go gets confused in the early days of the whole go modules thing, and I'm just a bit scarred by those memories.

I'll add go.mod later as I added some dependencies in #154

PR #154, opened 2020-04-13
Refactor JWS API

lestrrat:
fixes #149

# what's new

* API to iterate all headers pairs in a header
* API to access common headers (e.g. `h.Algorithm()`, `h.KeyID()`, etc.)

# what's changed

* Public fields have all been changed to private fields
* Types which didn't need to be public have been either removed or unexported
* `jws.Header` is now an interface. `jws.StandardHeader` no longer exists
* `jwt` Iterator API has been changed to conform with the new API for `jws`

PR #156, opened 2020-04-15
Refactor JWE API

lestrrat:
# What's changed

* JWE headers are no longer structs, but they are interfaces
* Parameters in a header are accessed via accessors, and they are generally immutable
* Headers have iterators and other utility functions, just like jws does after #154
* Fixes ECDHES decrypting
* Unexport a bunch of internal-only structs and functions


lestrrat:
I'm now done refactoring the package.

While I believe virtually nobody will have a problem, this requires API changes. am again letting it stew for a couple of days.

lestrrat:
I think this works. merging

PR #166, opened 2020-04-17
Add golangci-lint to CI

sonatard:
I disable any rules in https://github.com/lestrrat-go/jwx/pull/166/commits/b75330f749e8dca19f5cd6b926181a50a62e2d32. if you think need it, I'll fix it.

And `enable-all: true` is so strong rule. If you don't like it please tell me what disable rules.


jwe/internal/cmd/genheader/main.go
```diff
@@ -3,14 +3,15 @@ package main
...
 	"bytes"
 	"fmt"
-	"go/format"
 	"log"
 	"os"
 	"sort"
 	"strconv"
 	"strings"
 
+	"github.com/lestrrat-go/jwx/jwe"
```
lestrrat, line 12: no, don't import from the package that we're generating into. I know this causes problems

jws/internal/cmd/genheader/main.go
```diff
@@ -3,14 +3,15 @@ package main
...
 	"bytes"
 	"fmt"
-	"go/format"
 	"log"
 	"os"
 	"sort"
 	"strconv"
 	"strings"
 
+	"github.com/lestrrat-go/jwx/jws"
```
lestrrat, line 12: same here. no importing from the package that we're generating

sonatard:
Thanks for review! I fixed it.

lestrrat:
Thank you very much!

PR #195, opened 2020-06-21
fix bug where invalid algorithm results in nil error response

### LINKED ISSUE #194: Providing a bad algorithm to jwt.Sign should error, but does not.
**Describe the bug**

Calling this code should error, but doesn't:

```go
signed, err := jwt.Sign(tok, jwa.SignatureAlgorithm("BOGUS"), priv)
```

The bug is here: https://github.com/lestrrat-go/jwx/blob/9143661f037e46a9d2b5861fc19c4b051966b00b/jwt/jwt.go#L89

This never assigns the return value of `hdr.Set` to `err`, and so the `err` that's wrapped in the following line is `nil`, and `errors.Wrap` will return `nil` if the `err` is `nil`.

(above)

I've got a PR incoming, but wanted to issue a bug report first for context.

hairyhenderson:
Fixes #194 

Signed-off-by: Dave Henderson <dhenderson@gmail.com>

### DIFF
--- a/jwt/jwt.go
+++ b/jwt/jwt.go
@@ -86,10 +86,10 @@ func Sign(t Token, method jwa.SignatureAlgorithm, key interface{}) ([]byte, erro
 	}
 
 	hdr := jws.NewHeaders()
-	if hdr.Set(`alg`, method.String()) != nil {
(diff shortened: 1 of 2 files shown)


lestrrat:
LGTM. Thank you!

PR #197, opened 2020-07-02
Create a way to automatically choose a key to use for JWT verification from a given jwk.Set

### LINKED ISSUE #196: Allow JWKS-style sources of keys to verify JWT
The list of keys used to sign JWT may vary over time for a particular identity provider. They provide these at some URL which you're supposed to refresh "every so often", and supply some number of keys, identified by a "kid". This kid is also part of the JWT's header, and you're supposed to verify the JWT signature using the corresponding key. This is all implemented already in jwk.Set.

The rub is that you don't know which key to verify with prior to parsing the header of the JWT.

It's possible to use the Parse function directly as-is, and then separately perform a verify step, but this can be a bit more error prone. Perhaps it'd be possible to define a new WithVerify type of thing which integrates with a jwk.Set.

lestrrat:
fixes #196

### DIFF
--- a/jwt/jwt.go
+++ b/jwt/jwt.go
@@ -113,11 +113,11 @@ func lookupMatchingKey(data []byte, keyset *jwk.Set) (jwa.SignatureAlgorithm, in
 	}
 
 	var rawKey interface{}
-	if err := keys[0].Raw(rawKey); err != nil {
+	if err := keys[0].Raw(&rawKey); err != nil {
 		return "", nil, errors.Wrapf(err, `failed to construct raw key from keyset (key ID=%#v)`, kid)
 	}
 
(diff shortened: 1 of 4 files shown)


jwt/jwt.go
```diff
@@ -29,63 +30,135 @@ func ParseBytes(s []byte, options ...Option) (Token, error) {
 // The token must be encoded in either JSON format or compact format.
 //
 // If the token is signed and you want to verify the payload, you must
-// pass the jwt.WithVerify(alg, key) option. If you do not specify these
-// parameters, no verification will be performed.
+// pass the jwt.WithVerify(alg, key) or jwt.WithVerifyKeySet(jwk.KeySet) option.
```
imirkin, line 33: jwk.Set is the typename

lestrrat: Oh oops. nice catch

PR #198, opened 2020-07-07
Sign a token directly with a JWT key

tfaller:
Hello, 

I used this module today for the first time.
One thing was confusing. Why can't I sign a token directly with an given JWT Key?
If you have a valid Key (from a private KeySet) it should be straight forward to us this key to sing a token.
This pull request adds a new "SignWithKey" function that makes that easy.

Regards

### DIFF
--- a/jwt/jwt.go
+++ b/jwt/jwt.go
@@ -172,3 +172,31 @@ func Sign(t Token, method jwa.SignatureAlgorithm, key interface{}, options ...Op
 
 	return sign, nil
 }
+
+// SignWithKey creates a singed JWT token which is singed by the given JWT key
+func SignWithKey(t Token, key jwk.Key, options ...Option) ([]byte, error) {
+	var hdr jws.Headers
+	for _, o := range options {
+		switch o.Name() {
+		case optkeyHeaders:
+			hdr = o.Value().(jws.Headers)
+		}
+	}
+	if hdr == nil {
+		hdr = jws.NewHeaders()
+	}
+
+	kid := key.KeyID()
+	if kid != "" {
+		if err := hdr.Set(jwk.KeyIDKey, kid); err != nil {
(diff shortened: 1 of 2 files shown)


lestrrat:
I hear you, but instead of adding a new API, I think allowing `jws.Sign` to accept a `jwk.Key` along side other keys should suffice. Will look into it.

lestrrat:
#199  doesn't do 100% of what you initially proposed, but I think it's a better compromise than adding a new API.
In this PR. we added the ability to sign using JWK keys via `jws.Sign`. `jwt.Sign` still requires that we specify the signature
algorithm explicitly - which may be a bit annoying, but JWK keys can be passed directly through to `jws.Sign`, cutting your work of having to get the raw key and assign the key ID.

So in #199 you can do this:

```go
var jwkKey jwk.Key
...
jwt.Sign(token, jwa.SignatureAlgorithm(jwkKey.Algorithm()), jwkKey)
```

Just a s note, there is a historical reason and a practical reason to requiring the signature algorithm to be explicitly specified in the various `Sign()` functions. The practical reason is that library users tend to get these values straight out of forms, databases, and other data sources, and there was a more-than-you-expect fraction of cases where the algorithm/key pair was not supposed to match. The historical reason is that at least some of the previous libraries that I based this library's API seemed to also force the library users to explicitly specify the algorithm, and so with my previous observation and the precedents set by others, I followed this path. This is not to say that it's not a good idea to change the API, but to state why it's the way it is.

tfaller:
Thank you for your response and explenations.

I actually thought about this as well. I didn't because of, as you said, specifing the algorithm can be annoying.

But yes, I agree, doing it your way seems fine. My first attempt as a module user even was to use the Sign method with a JWK key and a provided  algorithm. So your way is pretty self explanatory.

lestrrat:
Cool, will merge and what not. Thanks!

PR #202, opened 2020-08-14
Make sure to use openid.Token when WithOpenIDClaims() is specified

### LINKED ISSUE #201: jwt.ParseBytes(buf, jwt.WithOpenIDClaims()) not work
jwt.ParseBytes, WithOpenIDClaims not work,  the returned token is not openid.Token type.

as you can find in jwt.go, the Parse() just ignores WithOpenIDClaims option
`func Parse(src io.Reader, options ...Option) (Token, error) {
	var params VerifyParameters
	var keyset *jwk.Set
	for _, o := range options {
		switch o.Name() {
		case optkeyVerify:
			params = o.Value().(VerifyParameters)
		case optkeyKeySet:
			keyset = o.Value().(*jwk.Set)
		}
	}
`

lestrrat:
fixes #201

### DIFF
--- a/jwt/jwt.go
+++ b/jwt/jwt.go
@@ -35,12 +35,15 @@ func ParseBytes(s []byte, options ...Option) (Token, error) {
 func Parse(src io.Reader, options ...Option) (Token, error) {
 	var params VerifyParameters
 	var keyset *jwk.Set
+	var token Token
 	for _, o := range options {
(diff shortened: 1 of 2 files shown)


lestrrat:
The test failure is a failure to upload to codecov, and we're just fine and dandy without it.

Issue #203, opened 2020-09-22
There should be jwt.ParseOption and jwt.VerifyOption


wxdao:
I find it really hard for me at first to understand `jwt.Parse()` only accepts `jwt.WithVerify` & `jwt.WithKeySet` and the rest like `jwt.WithAudience` are all for `token.Verify()`. How am I supposed to know what in those `jwt.Option`s  to be used? Furthermore, naming of `token.Verify()` and `jwt.WithVerify` is a bit confusing. I failed to use them correctly until I jumped into the definitions to actually read the code. Maybe something like `jwt.WithSingleKey` and `token.CheckStandardClaims` could be better?

lestrrat:
Let me break down your comment:

1. How am I supposed to know what in those jwt.Options to be used ....

Documentation. But yes, I can see they could use a lot of improvement. If you have concrete suggestions, please let me know (especially if you expected the definitions to be at one place and couldn't find them there)

2. Naming of token.Verify() and jwt.WithVerify is a bit confusing...

I don't have an opinion there, so if you have a concrete proposal, I can certainly consider it. Please create a pseudocode test case that shows how one would use (or not use) your options.

wxdao:
Thanks for your reply. I will file some MRs soon to better represent my ideas. However I'm not sure how much api change could still happen here since it's already into 1.0 version.

lestrrat:
Adding options should be doable. Changing existing calling conventions not so much. However, it's a proposal so you shouldn't haven to worry too much about backward compat (yet :)

PR #204, opened 2020-09-23
[WIP] Use distinct option types to allow compile-time check of correct usage

wxdao:
#203 

Methods like `jwt.Parse`, `jwt.Sign` and `jwt.Verify` use the same type alias of option.Interface. It's really hard to tell what in those `Option`s to be used for a method, since you can put any one in function calls and compile successfully, unless one digging into document and source code to find the correct usage. People are lazy, however :D. A compile-time check would greatly improve coding experience.

At first I tried to use type alias and type definitions reusing `option.Interface` to minimize the changes, such as:

```
type ParseOption = option.Interface
type VerifyOption = option.Interface
type SignOption = option.Interface
```

and:

```
type ParseOption option.Interface
type VerifyOption option.Interface
type SignOption option.Interface
```

No success however, go still considered these type are [assignable](https://golang.org/ref/spec#Assignability) allowing mixed use of them in function calls.

So I have to take a different approach to completely rewrite Option types with [functional options](https://github.com/tmrts/go-patterns/blob/master/idiom/functional-options.md). I'm not sure if this change is too heavy?

Currently `Option`s in jwt package is changed for demo. If we are happy about it I can continue to do the refactoring for the rest of packages.


lestrrat:
Thank you for the PR. interesting!


First off, let's just agree to keep this in `jwt` for the time being :) As you say, we may or may not adopt this yet.

---

So let me think + share with you why this interface is the way it is. The original reason I went with this option interface is because it was generally easier to generate code that has this signature.

For example, if we pass a config struct, we need to remember the correct incantation including namespaces and what not:

```go
fmt.Printf("¥n&foopkg.Config{") // need to remember this
if option1 {
  fmt.Printf("¥n  Option1: ....")
}
if option2 {
  fmt.Printf("¥n  Option2: ....")
}
fmt.Printf("¥n},") // need to remember to close
```

Whereas with a signature like `options ...Option`, you can just include everything in a loop:

```go
if option1 {
   options = append(options, `WithOption1(...)`)
}
if option2 {
   options = append(options, `WithOption2(...)`)
}
   fmt.Printf("¥n%s,", option)
}
```

Also, I always found config objects to be bulky and messy. For one thing I don't like public struct fields in most cases. This is just a personal preference, but... I also don't like the fact that it makes it easy to carry said config object around. I have a feeling config objects, if used, should be ephemeral, but it's hard to enforce it.

---

Anyways. So I think your proposal does in fact do not affect what I wanted to do, so I guess I'm okay with it in principle.

What I don't like in the current form of PR is the fact that the there's an action-at-a-distance thing going on with regards to the config objects and the options. From what I can see, I think the options should be data, and the config objects should be the collectors.

But at the same time I understand that putting this logic in the config object also has its drawbacks, mainly in that it's again hard to tell what the difference in options types are.

---

So in principle I'm ok with the change, but I currently have doubts in the implementation. Feel free to send me more refinements, but for the time being I'm going to be sitting on it and pondering for a bit.

(Also, disclaimer: I'm currently a bit swamped at work, so I don't know how fast I can get back to this)

wxdao:
Thanks for the details!

What about the options pattern in mongo-go-driver, like [this](https://github.com/mongodb/mongo-go-driver/blob/master/mongo/options/aggregateoptions.go#L55). It has option data and option methods together.

Code will be like:

```
type ParseOptions struct {
    keySet *jwk.Set
    // ...
}

func (o * ParseOptions) SetKeySet(keySet *Set) {
    o.keySet = keySet
}

// ...

func ParseString(s string, opt *ParseOptions) {
    // make use of `opt.keySet` etc.
}

// usage
func main() {
    jwt.ParseString("some token", jwt.ParseOptions().SetKeySet(keySet).SetOpenIDToken())
}
```

However it will cause significant api change. No `WithXXX` anymore.

lestrrat:
Right the fact that the mongo way completely changes the API means that even if we decide that that's the way to go we can't really change it any time soon. The only workaround I can think of is to introduce a different set of `Parse()` functions like `Parse2(mongoStyleOptions)`, which, coming from a C background is not completely out, but yeah, it looks horrible :)

---

Let me step back again. The reason the `Option` interface was public was because it allowed external users to specify arbitrary options variable should the need arise. I dunno, for example, maybe we had a bug in our `Option` object, and used a different key. Then the user could potentially provide their own `Option` object:

```go
// pretend mypkg.CorrectOption fulfills the `Option` interface
jwt.Parse(.... , &mypkg.CorrectOption{})
```

But if we choose not to allow this, we could potentially do this:

```go
type ParseOption interface {
   Name() string
   Value() interface{}
   isParseOption()
}

type parseOptionMarker struct{}
func (m parseOptionMarker) isParseOption() {}

type parseOption struct {
    parseOptionMarker
    name string
    value  interface{}
}

func WithBlah(...) ParseOption {
   return &parseOption{ name: blah, value ... }
}

func WithBoo(...) ParseOption {
  return &parseOption{ name: boo, value ... }
}
```
Food for thought

lestrrat:
Hmm, I'm working on a separate project where I need to work on options for work, and I think I like the use-unexported-method-in-interface method to provide type safety. This way the implementation doesn't necessary need to change, but we _can_ differentiate between the option types. Thoughts?

wxdao:
Yes, I agree your use-unexported-method-in-interface approach would better fit the current codebase, with minimal changes needed.

I will find some time to update the MR with that if this is we all are happy about, but I've got some thoughts about what you said earlier.

1) Your reason about `option.Interface` being public.

You mention that this design allow users to specify arbitrary options variable should the need arise. I think this is ridiculous and impossible should there be a need for this. Tho `option.Interface` is public, the actual config keys are internal and undocumented. Only WithXXX option methods are documented and they are the only options(no pun intended) a sane user will discover and rely on.

A bug should never happen on WithXXX methods. If there is, it's a bug of the corresponding option accepting methods, unless some action-at-a-distance causes it.

2) And speaking of action-at-a-distance, I think there's no difference before the refactor.

Look at this original code:

```
const (
 	optkeyAcceptableSkew = "acceptableSkew"
 	optkeyClock          = "clock"
 	optkeyIssuer         = "issuer"
 	optkeySubject        = "subject"
 	optkeyAudience       = "audience"
 	optkeyJwtid          = "jwtid"
 )

func WithClock(c Clock) Option {
 	return option.New(optkeyClock, c)
 }
```

Does it really distinguish itself from the below after the refactor?

```
type verifyOptions struct {
 	issuer      string
 	subject     string
 	audience    string
 	jwtid       string
 	clock       Clock
 	skew        time.Duration
 	claimValues map[string]interface{}
 }

func WithClock(c Clock) VerifyOption {
 	return func(vo *verifyOptions) {
 		vo.clock = c
 	}
 }
```

You still have to define option keys besides the actual method, or you will need to use string literals, while every linter will warn about this. The later approach also helps defining the actual config item type instead of `interface{}` everywhere, I see it better.

lestrrat:
this is mainly just for record: for (1) yeah, it was a defensive instinct: I didn't want to corner myself in a spot where only the author has control over what options are being passed, but it was probably doesn't make sense. For (2) yes, I think it's action at a distance, because instead of the option object being just a data carrier, and the consumer being the one decides exactly what to do with the data, with your approach requires that both the consumer and the option object needing to coordinate how to use the config object.

wxdao:
Hi, I just decide to roll my own library and stop investing more time in this. I apologize my sudden getting involved and then abandoning. Thank you for your great work here.

PR #206, opened 2020-09-28
Add missing PrivateParams method

lestrrat:


lestrrat:
The failures here are limited to go-tip, so I'm going to punt fixing this (I mean, the errors look weird -- why would go generate need to download an external dependency?), and merge

PR #208, opened 2020-09-29
Use a padding so that the buffer always aligns with the bit size

lestrrat:


jwk/ecdsa.go
```diff
@@ -157,25 +158,86 @@ func (k ecdsaPublicKey) Thumbprint(hash crypto.Hash) ([]byte, error) {
...
-		base64.EncodeToString(key.X.Bytes()),
-		base64.EncodeToString(key.Y.Bytes()),
+		base64.EncodeToString(xbuf),
+		base64.EncodeToString(ybuf),
 	), nil
 }
 
+var crvFixedBufferPool = sync.Pool{
+	New: func() interface{} {
+		return make([]byte, 66)
```
anatol: It might be a good idea to use some readable const with explanation for this `66` value.

lestrrat: Yeah, I was going meh, yeah, nah, etc about it. I guess I might as well.

Issue #207, opened 2020-09-28
Incorrect webkey thumbprint calculation


anatol:
Hi

I am porting some `jwk` functionality to `jwx` v 1.0.5 library (currently the latest release).

I found an EC thumbprint calculation incompatibility with `go-jose` and C `jose` libraries. Here is a webkey:

`{"kty":"EC","alg":"ECMR","crv":"P-521","key_ops":["deriveKey"],"x":"AJwCS845x9VljR-fcrN2WMzIJHDYuLmFShhyu8ci14rmi2DMFp8txIvaxG8n7ZcODeKIs1EO4E_Bldm_pxxs8cUn","y":"ASjz754cIQHPJObihPV8D7vVNfjp_nuwP76PtbLwUkqTk9J1mzCDKM3VADEk-Z1tP-DHiwib6If8jxnb_FjNkiLJ"}`

C `jose` library gives following thumbprint:

```shell
jose jwk thp -i- <<< '{"kty":"EC","alg":"ECMR","crv":"P-521","key_ops":["deriveKey"],"x":"AJwCS845x9VljR-fcrN2WMzIJHDYuLmFShhyu8ci14rmi2DMFp8txIvaxG8n7ZcODeKIs1EO4E_Bldm_pxxs8cUn","y":"ASjz754cIQHPJObihPV8D7vVNfjp_nuwP76PtbLwUkqTk9J1mzCDKM3VADEk-Z1tP-DHiwib6If8jxnb_FjNkiLJ"}'
2Mc_43O_BOrOJTNrGX7uJ6JsIYE
```

But `jxk` gives `ZNIw9ANrxjuqoiqZjzJD1pGr43w` thumbprint.

I started comparing how the values are computed in `jwk` vs `go-jose` and found that `go-jose` performs `x` and `y` padding to the size of the EC curve. Such padding adds one extra `0` in the front of `key.X` and `key.Y`.

As a proof of concept I ported code from `go-jose` and with the patch below the thumbprints become identical between different implementation:

```patch
diff --git a/jwk/ecdsa.go b/jwk/ecdsa.go
index 61736b0..272eacb 100644
--- a/jwk/ecdsa.go
+++ b/jwk/ecdsa.go
@@ -157,11 +157,14 @@ func (k ecdsaPublicKey) Thumbprint(hash crypto.Hash) ([]byte, error) {
        if err := k.Raw(&key); err != nil {
[...]

PR #209, opened 2020-10-01
ECDHS fix

lestrrat:
from #207 

* [x] Implement flattened JSON serialization  format.
* [x] Fix weird error with key unwrapping

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: [ 'tip', '1.15', '1.14' ]
+        go: [ '1.15', '1.14' ]
     name: Go ${{ matrix.go }} test
     steps:
       - name: Checkout repository
--- a/internal/concatkdf/concatkdf.go
+++ b/internal/concatkdf/concatkdf.go
@@ -3,7 +3,6 @@ package concatkdf
 import (
 	"crypto"
 	"encoding/binary"
-	"hash"
 
 	"github.com/lestrrat-go/jwx/buffer"
 	"github.com/pkg/errors"
@@ -11,9 +10,8 @@ import (
 
 type KDF struct {
 	buf       []byte
-	hash      hash.Hash
+	hash      crypto.Hash
 	otherinfo []byte
-	round     uint32
 	z         []byte
 }
 
@@ -30,19 +28,20 @@ func New(hash crypto.Hash, alg, Z, apu, apv, pubinfo, privinfo []byte) *KDF {
 	copy(concat[n:], privinfo)
 
 	return &KDF{
-		hash:      hash.New(),
+		hash:      hash,
 		otherinfo: concat,
-		round:     1,
 		z:         Z,
 	}
 }
 
-func (k *KDF) Read(buf []byte) (int, error) {
-	h := k.hash
-	for len(buf) > len(k.buf) {
+func (k *KDF) Read(out []byte) (int, error) {
+	var round uint32 = 1
+	h := k.hash.New()
+
+	for len(out) > len(k.buf) {
 		h.Reset()
 
-		if err := binary.Write(h, binary.BigEndian, k.round); err != nil {
+		if err := binary.Write(h, binary.BigEndian, round); err != nil {
 			return 0, errors.Wrap(err, "failed to write round using kdf")
 		}
 		if _, err := h.Write(k.z); err != nil {
@@ -53,10 +52,10 @@ func (k *KDF) Read(buf []byte) (int, error) {
 		}
 
 		k.buf = append(k.buf, h.Sum(nil)...)
-		k.round++
+		round++
 	}
 
-	n := copy(buf, k.buf[:len(buf)])
-	k.buf = k.buf[len(buf):]
+	n := copy(out, k.buf[:len(out)])
+	k.buf = k.buf[len(out):]
 	return n, nil
 }
--- a/internal/concatkdf/concatkdf.go
+++ b/internal/concatkdf/concatkdf.go
@@ -5,6 +5,7 @@ import (
 	"encoding/binary"
 
 	"github.com/lestrrat-go/jwx/buffer"
+	"github.com/lestrrat-go/pdebug"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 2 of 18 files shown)


lestrrat:
@anatol I got this far today, but I'm not sure how much more time I'm going to have within the next few days. I suspect I'm passing something wrong to `ECDHESDecrypt.Decrypt`, but I haven't figured it out yet.

anatol:
At quick glance I see a few differences from `jose` library.

1. 
```go
kdf := concatkdf.New(crypto.SHA256, []byte(kw.algorithm.String()), z.Bytes(), kw.apu, kw.apv, pubinfo, []byte{})
```

This line uses `kw.algorithm.String()` as a second argument. While `jose` uses content encryption name here (such as `A256GCM` from msg.protectedHeader.ContentEncryption).

2.
```go
block, err := aes.NewCipher(kek)
if err != nil {
        return nil, errors.Wrap(err, "failed to create cipher for ECDH-ES key wrap")
}

return Unwrap(block, enckey)
```

this part is only used for algorithms with key wrapping (i.e. the ones that have suffix like `+A256KW`). For plain `ECDH-ES` algorithm is should be ommited and use `kek` as content encryption key. It also means that in case of `ECDH-ES` `keysize` will depend on `contentEncryption` algorithm rather than on key wrapping algorithm.

cc @sergio-correia @npmccallum

lestrrat:
thanks for your input. I've yet to finish this but I see where I was going wrong. working on it when in between $day_job...

anatol:
Thank you very much for your great work @lestrrat!

lestrrat:
@anatol well, I finally got around to fixing things. Care to take it for a spin?

anatol:
Thank you for your patches. Things look better now. `ECDH` key decryption looks correct.

Though message decryption still fails in the line 
```go
plaintext, err = cipher.Decrypt(cek, iv, ciphertext, tag, computedAad)
```
It is interesting that with current `master` branch this line (when using the same `cek`) works fine.

It turns out that the current topic branch has changed `aad` value. Previously it was (decoded):

```json
{"alg":"ECDH-ES","clevis":{"pin":"tang","tang":{"adv":{"keys":[{"alg":"ES512","crv":"P-521","key_ops":["verify"],"kty":"EC","x":"AGeZ7xh05hGfO8HEaDYCr855kq8FEw5Hxzo1Ne30A7frqQq0u_hbn_Vo9lUaLAwce8o2a-soEKGa7pwmIeFUm7Io","y":"AEUeuTA_Ub4u605OGkDrpQHFCqXhIjPhTZAIVFIrdpU-iEQ_7rzyBwH2OQJMvScmkRk5UtfHxnUskbHU7DIzX89C"},{"alg":"ECMR","crv":"P-521","key_ops":["deriveKey"],"kty":"EC","x":"AJwCS845x9VljR-fcrN2WMzIJHDYuLmFShhyu8ci14rmi2DMFp8txIvaxG8n7ZcODeKIs1EO4E_Bldm_pxxs8cUn","y":"ASjz754cIQHPJObihPV8D7vVNfjp_nuwP76PtbLwUkqTk9J1mzCDKM3VADEk-Z1tP-DHiwib6If8jxnb_FjNkiLJ"}]},"url":"http://localhost"}},"enc":"A256GCM","epk":{"crv":"P-521","kty":"EC","x":"AQ6QNb4mtocsleCSojjhI-0AAViPpGk0fzH0C1Q2d90I2nA5k9Ig6dGe40u_c0sjxoEV3IiYA_hiWd_pi3kov7Xn","y":"AFf-mPYwhd4BHre-6-JOclBeDW9c-kXQqn30bLQiX4Lmp6SMPM_BLleWbhAPR91haVDhiqsPWJF1iVAcOVMQX-P9"},"kid":"2Mc_43O_BOrOJTNrGX7uJ6JsIYE"}
```

now it fails and `aad` value is
```json
{"alg":"ECDH-ES","enc":"A256GCM","epk":{"kty":"EC","crv":"P-521","x":"AQ6QNb4mtocsleCSojjhI-0AAViPpGk0fzH0C1Q2d90I2nA5k9Ig6dGe40u_c0sjxoEV3IiYA_hiWd_pi3kov7Xn","y":"AFf-mPYwhd4BHre-6-JOclBeDW9c-kXQqn30bLQiX4Lmp6SMPM_BLleWbhAPR91haVDhiqsPWJF1iVAcOVMQX-P9"},"kid":"2Mc_43O_BOrOJTNrGX7uJ6JsIYE","clevis":{"pin":"tang","tang":{"adv":{"keys":[{"alg":"ES512","crv":"P-521","key_ops":["verify"],"kty":"EC","x":"AGeZ7xh05hGfO8HEaDYCr855kq8FEw5Hxzo1Ne30A7frqQq0u_hbn_Vo9lUaLAwce8o2a-soEKGa7pwmIeFUm7Io","y":"AEUeuTA_Ub4u605OGkDrpQHFCqXhIjPhTZAIVFIrdpU-iEQ_7rzyBwH2OQJMvScmkRk5UtfHxnUskbHU7DIzX89C"},{"alg":"ECMR","crv":"P-521","key_ops":["deriveKey"],"kty":"EC","x":"AJwCS845x9VljR-fcrN2WMzIJHDYuLmFShhyu8ci14rmi2DMFp8txIvaxG8n7ZcODeKIs1EO4E_Bldm_pxxs8cUn","y":"ASjz754cIQHPJObihPV8D7vVNfjp_nuwP76PtbLwUkqTk9J1mzCDKM3VADEk-Z1tP-DHiwib6If8jxnb_FjNkiLJ"}]},"url":"http://localhost"}}}
```

I see that the new `aad` json has its key non-ordered. I assume the keys have to be ordered.

lestrrat:
Those look like new test cases for me (which is fine, but please give me the test cases! :D

As for the aad,  I did find that the key ordering is different from go-jose. I was grepping through RFC 7516 for words like sort/order but I can seem to find this in the spec, so I had punted it. 

go-jose uses json.Masrhal, which IIRC sorts the keys of a map, but unless it's in the RFC, I think it's just a coincidence? Let me know if you know this is not true.

If the spec states that the keys must be ordered lexicographically, I will happily do this. One thing though, I should at least sort the private parameters -- all the predefined keys like "alg", "enc" will be in stable order because they come from a struct, but I need to fix the private parameters myself.

lestrrat:
on second thought maybe I'll just bite bullet and sort everything....

anatol:
> Those look like new test cases for me (which is fine, but please give me the test cases! :D

A test case for it would be an encrypted message with an element in `privateParam`. In my case I use `clevis` generated message with `ECMR` key derivation algorithm. I have implemented it here https://github.com/anatol/jwx/commit/6b461f12b2552791d627f0d91f921ed18635c5d9

There should be a way to generate similar encrypted message for `ECDH`.

As of ordering - `aad` content is used for authenticated decryption. So there has to be **some** stable order of the serialized fields.

cc @sergio-correia and @npmccallum who have better knowledge of this area.

lestrrat:
> As of ordering - aad content is used for authenticated decryption. So there has to be some stable order of the serialized fields.

Yeah, I wrote my previous comment and I was immediately thinking: But what I just wrote is nonsense. So I'm going to add ordering, but I just find it weird it's not explicitly stated in the RFC, or maybe I'm just blind.

lestrrat:
@anatol Added header ordering (which in turn should make unencrypted aad contents ordered) in 2ea01d3 . I haven't written tests yet, but that's all the tuits I have for today. Crossing my fingers.

anatol:
With 2ea01d3 my integration tests that involve `jose` `clevis` and `jwx` finally pass. Woot! Thank you for your work.

lestrrat:
cool, I'm going to do some merging work in my spare time.

PR #213, opened 2020-10-08
Implement direct key encoding

anatol:
This key encryption mode simply bypasses a key provided by the user to
the content encryptor.

Here is an example of usage:

  key := []byte{...}
  jwe.Decrypt(data, jwa.DIRECT, key)

Signed-off-by: Anatol Pomozov <anatol.pomozov@gmail.com>


lestrrat:
Thank you!

PR #214, opened 2020-10-13
Allow jwt.Parse using JWKS to use default single key when no kid in JWT header

darnmason:
I believe it's standard practice that if a JWKS only contains a single key then the kid isn't necessarily required in the JWT header. The validation code can assume it should use the only available key.

However, I haven't read the spec so if this is not the case then please disregard this PR and I'll continue to use my fork.

To keep the API the same and not change existing behaviour I added a new option for jwt.Parse to allow this behaviour.

- If the useDefault flag is true and there's a single key in the set it'll use that to verify the token.
- If the useDefault flag is true and there's multiple keys in the set it'll error.
- If the useDefault flag is false then the behaviour is unchanged.

I wasn't sure on naming, **_With..._** didn't seem to fit the use case here, so I went with **_UseDefaultKey_**.

### DIFF
--- a/jwt/jwt.go
+++ b/jwt/jwt.go
@@ -35,6 +35,7 @@ func ParseBytes(s []byte, options ...Option) (Token, error) {
 func Parse(src io.Reader, options ...Option) (Token, error) {
 	var params VerifyParameters
 	var keyset *jwk.Set
+	var useDefault bool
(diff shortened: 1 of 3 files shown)


lestrrat:
Sounds like a plan. Thank you!

Issue #211, opened 2020-10-07
Add integration tests with use of `jose` tool


anatol:
One thing that would help keeping `jwx` feature-full and compatible with other implementation is to have integration tests that utilize those implementation. The idea here is to generate a key using tools like `jose`, then encrypt a message (again using `jose`) and then try to decrypt it with `jwx`. At the end the plaintext should be exactly the same as input to `jose`.

The test should also verify data flow other way - encrypt a text with `jwx` and then it decrypt using `jose` tool. The end result should match input to `jwx`.

lestrrat:
just letting you know that this hasn't been warnocked. been taking time off for the last week. Will get back when I sort out $day_job :)

anatol:
No worries.

Just a bit more info about this request. We want to run some operation with existing tools like `jose` and then check the result using `jwx`. e.g

```
jose jwk gen -i '{"alg":"A128GCM"}' -o key.jwk
jose jwk thp -i key.jwk
jose jwe enc -I msg.txt -k key.jwk -o msg.jwe
```

generates a key, generates a thumbprint and then encrypts a message. `jwx` should read the key, parse it, check that thumbprint matches, decrypt the message.

We want different types of keys here and check different operations (encrypt/decrypt/sign/verify/..)

And on top of that we want to test the functionality the other way - generate a key and encrypt a message using `jwx` and then decrypt it with `jose`. It will give a pretty good compliance coverage for `jwx`.

lestrrat:
@anatol So I added some basic testing for `jose` tools in #216 

Maybe you can suggest some tests that you want me to include in there?

lestrrat:
closing this as at least some tests have been added.

PR #216, opened 2020-10-19
add jose interop tests

lestrrat:
refs #211

### DIFF
--- a/jwx_test.go
+++ b/jwx_test.go
@@ -183,13 +183,13 @@ func TestJoseCompatibility(t *testing.T) {
 
 			expected := []byte("hi")
 
-			cryptfile, jwecleanup, err := jose.EncryptJwe(t, ctx, expected, keyfile)
+			cryptfile, jwecleanup, err := jose.EncryptJwe(ctx, t, expected, keyfile)
 			if !assert.NoError(t, err, `jose.EncryptJwe should succeed`) {
 				return
 			}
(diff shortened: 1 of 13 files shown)


lestrrat:
9199de3 had already been merged, but a I screwed up the order of operation between push and PR

lestrrat:
meh, I can't find the jose binary in Github Actions, but I'm out of tuits.

jwx_test.go
```diff
@@ -0,0 +1,141 @@
...
+			expected := []byte("hi")
+
+			cryptfile, jwecleanup, err := jose.EncryptJwe(t, ctx, expected, keyfile)
+			if !assert.NoError(t, err, `jose.EncryptJwe should succeed`) {
+				return
+			}
+			defer jwecleanup()
+
+
+			payload, err := jose.DecryptJwe(t, ctx, cryptfile, keyfile)
```
anatol: this test generates a key ancrypts and decrypts with `jose` tool. It would be more useful to add codepath that exercises `jwx` path here. e.g.

1. generate a key using `jwx`
2. encrypt a file using `jose`
3. decrypt it using `jwx`

and then swap 2 with 3

lestrrat: ha ha you are absolutely right.
I was probably thinking "I need to complete this tool's API!" and ust ran with it

PR #222, opened 2020-10-28
Add a way to configure the global behavior of json Marshaling

### LINKED ISSUE #221: JWS headers with numeric values are parsed as float64
When I parse a JWS into a Message and then access the headers, any number (e.g. a timestamp in my case, originally an int64) is unmarshaled as float64. This potentially loses precision and is not ideal for security code.

Setting the header:
```go
	_ = headers.Set("ts", time.Now().Unix()) 
```

Reading the header:
```go
	tsi, found := headers.Get("ts")
	ts, ok := tsi.(float64)
	if !found || !ok || !isValidTimestamp(int64(ts)) { // conversion to int64 will round the number :-(
		logger.Debugf("ts: found %v ok %v tsi %T -- %v", found, ok, tsi, tsi)
		return errors.New("Bad ts header")
	}
```

A better solution would be to use the JSON package `UseNumber` facility, then we'd get back a string. See https://stackoverflow.com/a/22346593

lestrrat:
fixes #221

Introduce a function to configure global behavior of json Marshal/Unmarshaling. Currently only supports `jwx.WithUseNumber` to either toggle `"encoding/json".Decoder.UseNumber` on or off. This flag has global
effect, so the user must be aware of its existence, and opt to accept its side effects in the entire program
that uses this.

```go

func init() {
  jwx.DecoderSettings(jwx.WithUseNumber(true))
}

... elsewhere ...

jws.Parse(...) // uses json.Number
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -370,6 +370,9 @@ func init()
 }
 ```
 
+Do be aware that this has *global* effect. All code that calls in to `encoding/json`
+within `jwx` *will* use your settings.
+
 ## Other related libraries:
 
 * https://github.com/dgrijalva/jwt-go
--- a/README.md
+++ b/README.md
@@ -359,6 +359,17 @@ Supported content encryption algorithm:
 
 PRs welcome to support missing algorithms!
 
+## Configuring JSON Parsing
+
+If you want to parse numbers in the incoming JSON objects as json.Number
+instead of floats, you can use the following call to globally affect the behavior of JSON parsing.
+
+```go
+func init()
+  jwx.DecoderSettings(jwx.WithUseNumber(true))
+}
+```
+
 ## Other related libraries:
 
 * https://github.com/dgrijalva/jwt-go
--- a/buffer/buffer.go
+++ b/buffer/buffer.go
@@ -6,7 +6,8 @@ package buffer
 import (
 	"encoding/base64"
 	"encoding/binary"
-	"encoding/json"
+
+	"github.com/lestrrat-go/jwx/internal/json"
 
 	"github.com/pkg/errors"
 )
--- a/internal/json/json.go
+++ b/internal/json/json.go
@@ -13,13 +13,14 @@ type RawMessage = json.RawMessage
 var muGlobalConfig sync.RWMutex
 var useNumber bool
 
+// Sets the global configuration for json decoding
 func DecoderSettings(inUseNumber bool) {
 	muGlobalConfig.Lock()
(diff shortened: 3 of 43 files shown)


yaronf (approved):
Thank you for this feature!

Quite frankly, I don't know how to run my existing code with the new branch. But I read your code and ran the tests, I understand what it's doing and I think it's a good solution for the problem. 

Another flag which might have been interesting is `DisallowUnknownFields` but when I hard-code it, it beaks the JWS test. So probably not a good idea.

lestrrat:
@yaronf re: DisallowUnknownFields, the whole parsing code for private parameters everywhere will have to change for apparently no gain. plus  I don't think I want to add a flag that I don't know if anybody is using it just for completeness' sake.

For this PR, I think it doesn't break anything for users who don't opt-in to it, so I'm content to merge it. Won't be doing a release now tho, since it's a bit late at night here.

yaronf:
Thank you again, after the merge I pulled the code and verified that it does work for my application (with and without the flag). Have a good night!

cloudkucooland:
Firstly, thanks for all your work on this. This package is fantastic.

I just upgraded to test these changes, and I'm now seeing 

RISC/GoogleRISC.go:159:19: cannot use "github.com/lestrrat-go/jwx/jwt".WithIssuer("https://accounts.google.com") (type "github.com/lestrrat-go/jwx/jwt".ValidateOption) as type "github.com/lestrrat-go/jwx/jwt".ParseOption in argument to "github.com/lestrrat-go/jwx/jwt".Parse:
	"github.com/lestrrat-go/jwx/jwt".ValidateOption does not implement "github.com/lestrrat-go/jwx/jwt".ParseOption (missing "github.com/lestrrat-go/jwx/jwt".isParseOption method)

What else do I need to tweak in this block?

		token, err = jwt.Parse(bytes.NewReader(rawjwt),
		  jwt.WithValidate(true),
		  jwt.WithVerify(jwa.RS256, &pk),
		  jwt.WithIssuer("https://accounts.google.com"),
		)

commenting out WithIssuer makes it compile just fine, but I'd really rather have that test in place.

lestrrat:
Ah nice catch. I forgot about that package.

lestrrat:
On second thought, this is from #220. Will review

lestrrat:
@cloudkucooland this being a fairly annoying issue, I'm going to merge #223 without you reviewing. Please holler at a separate issue if this is still a problem

PR #223, opened 2020-10-29
fix jwt parse options

lestrrat:


cloudkucooland:
The compile and initial tests were clean. Thanks for the quick turn-around on this.

I'll run more extensive tests in the next few days.

cloudkucooland:
Installed it on several systems, one was failing... turns out the clock on it was ~30 seconds off and it was failing on the iat test. After fixing the clock, everything is good.

The recent changes are improving functionality. Good work!

Issue #238, opened 2020-11-24
jws: inconsistent documentation on fine-grained verification


debug-ito:
Document of `jws` (v1.0.5) is inconsistent about fine-grained verification.

Document on `Verify` function says

> If you need more fine-grained control of the verification process, manually call `Parse`, generate a verifier, and call `Verify` on the parsed JWS message object.

However, document on `Message` struct says

> Do not expect to use the Message object to verify or construct a signed payloads with. You should only use this when you want to actually want to programmatically view the contents for the full JWS payload.

I suggest removing the comment on `Verify` function, because manually doing `Parse` and `Verify` is harder than it looks like. To do that, the user has to concatenate the JWS payload and the protected header manually, and pass the result to the `payload` argument of `Verify` method of `Verifier` interface. Although the parameter is named `payload`, it's not a JWS payload. The user has to read and understand RFC 7515 to do manual verification.

lestrrat:
first, I think the intention of the comment in Verify is that if you want to verify an arbitrary payload and signed string, you need to use the low level tools, such as the `verify` and `sign` package

second, the comment on the Message does not contradict anything from the previous Verify comment. It explicitly states that a message object is not meant to be used for signing. It's meant to be used when you want to view the contents of the signed payload. i.e., there's no such thing as `Message.Sign()`, but in order to know what `eyJ0eXAiOiJKV1QiLA0KICJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJqb2UiLA0KICJleHAiOjEzMDA4MTkzODAsDQogImh0dHA6Ly9leGFtcGxlLmNvbS9pc19yb290Ijp0cnVlfQ.dBjftJeZ4CVP-mB92K27uhbUJU1p1r_wW1gFWFOEjXk` contains in each section, you need to parse it down to `jws.Message`.

The wording could perhaps be better (PRs welcome), but I don't think they are wrong per se.

debug-ito:
Thanks for reply. I think I understand the spirit of those documents.

> call Verify on the parsed JWS message object.

In the above part of the document of `Verify` function, what does "the parsed JWS message object" refer to? Is that the `*jws.Message` value returned from `jws.Parse` function?

If the answer to the above question is yes, it seems to me that a `jws.Message` value is used for verification. So I thought it's inconsistent.

lestrrat:
[...]

PR #240, opened 2020-11-24
jws: fix document on Verify method

debug-ito:
Fix for #238.

`jws.Message` object is for inspecting JOSE headers and signatures in a JWS message (in receiver usecases). It's not meant to be passed to `Verify` method of `Verifier`.


lestrrat:
LGTM

Issue #243, opened 2020-11-27
X.509 certificates must be base64 encoded with padding


reinkrul:
According to RFC7515 section 4.1.6 (JWS, https://tools.ietf.org/html/rfc7515#section-4.1.6) and RFC7519 section 4.7 (JWK, https://tools.ietf.org/html/rfc7517#section-4.7) the X.509 certificate chain (`x5c`) must be encoded as base64 with padding. Since version 1.x the jwx library encodes the `x5c` field without padding, which is incorrect:

- Base64 RFC 4648 section 3.2 (https://tools.ietf.org/html/rfc4648#section-3.2) states that applications MUST use padding unless explicitly specified.
- the JWK and JWS RFC above don't specify omitting base64 padding for the `x5c` field.
- the `x5c` examples in the JWK and JWS RFCs all use padding.

It's important to note that the base64 encoding for compact serialization (URL-encoded base64) is without padding which is correct, because the RFC explictely states padding must be omitted.

Second important note: this library produced `x5c` fields _with_ padding pre-v1, it seems to be broken since v1. That causes JWK and JWS's produced in v1 (using the `x5c` field) not parseable in pre-v1 versions of this library.

If you agree this is a bug, I can supply a fix?

lestrrat:
I have only glanced through this issue and the rfc, but yeah, if base64 (not base64url) is required, I have a feeling I did screw that one up.

reinkrul:
I linked a PR that will fix the issue.

lestrrat:
@reinkrul Thanks!

PR #244, opened 2020-11-27
Encode x5c field as base64 with padding to conform with JWK and JWS RFCs

reinkrul:
Fixes https://github.com/lestrrat-go/jwx/issues/243


lestrrat:
LGTM

PR #245, opened 2020-12-13
Adding missing key encryption support - A*GCMKW, PBES2-*

imirkin:
I don't actually need these ... but seemed like fun to implement. Thanks for making this nice library. The only testing I've done is with the `jose` round-trip tests in `jwx_test.go`. Please feel free to suggest more.

I also don't love the duplication between decrypt and keyenc. Not sure why the symmetric ciphers got split out like that, but I tried to maintain the existing structure. Again - very open to suggestions for improvement.

Note that I had to add a dependency on `golang.org/x/crypto` for the `pbkdf2` logic. It's a fairly small function, so if the dependency is too heavy, we could copy it out as that lib is BSD-licensed.


lestrrat:
re: duplication between decrypt and keyenc, TBH I forget why and how much duplication exists as it is now, but I'm okay with shuffling things around. Please let me know if you have suggestions.

Just as a note, though: the original intention for the code organization was to move things around based on each component's responsibilities. I was deciphering other people's code and had a real hard time understanding where each functions came into play when everything was dumped into one place. 

everything else seems pretty legit. thank you.

PR #246, opened 2020-12-14
Improve jose interop logic to allow more tests (and some fixes)

imirkin:
The current interop bits for ECDH are useful everywhere. I made copies in my earlier PR, but a better approach is to just make it data-driven. If this is accepted, I'll rebase the other PR on top of this.

I was trying to fix ECDH-ES (unsuccessfully), but came across some fixes which seem right either way.


imirkin:
Note that the doubling change causes Test_A256KW_A256CBC_HS512 to break. However I'm not entirely sure why it wants to disallow this combination. Seems to work fine with jose.

    $ ./build/cmd/jose jwk gen -i '{"alg": "A256KW"}' -o tt
    $ echo foo | ./build/cmd/jose jwe enc -k tt -i '{"protected":{"enc":"A256CBC-HS512"}}' -I -
    {"ciphertext":"sybbfWmJF-ShlnGVi-jXNA","encrypted_key":"z9GDDlC-73JlLE5ObUGJoJ_0gmnEl2qBk6P-NoiJgjMakOx9dzp98K_8_XWPY6tbR6i4XixfPpDbXQCxcJ3UcNYQEf7siHPg","header":{"alg":"A256KW"},"iv":"82h8FWuK2QepdqjEwYOqRA","protected":"eyJlbmMiOiJBMjU2Q0JDLUhTNTEyIn0","tag":"SM8mqzyXIFroS7dbA-NF5yfhwkXeV57pg4nP34-IQmE"}

lestrrat:
Yeah, I'd rather do things data-drive myself, but I wasn't sure what kind of tests I was going to end up with, so I just took the straight forward way out.

Seems like after the merge I made for #245 there are conflicts. Would you kindly take a look at those?

imirkin:
Sure - in the meanwhile, any idea if I can get rid of the key size check? I don't see anything in the specs or in jose that would support those restrictions. I'm talking about this code:

```
		keysize = contentcrypt.KeySize()
		switch aesKeySize := keysize / 2; aesKeySize {
		case 16, 24, 32:
		default:
			return nil, errors.Errorf("unsupported keysize %d (from content encryption algorithm %s). consider using content encryption that uses 32, 48, or 64 byte keys", keysize, contentalg)
		}
```

Which was added in 9f294080eaf83cc1a6702ae31899021620d3bbb9 albeit without specific justification.

lestrrat:
I think I stole that from go-jose, but I admit I was going for compatibility without checking for its validity or necessity. I think if there's a good test that demonstrates that it's not required (note: I haven't fully reviewed your code, so it might already be there) then I think we're good

imirkin:
Well, I looked at PR #26 which had some more details. It calls out A256KW + A256CBC-HS512 as incompatible. However those are quite compatible -- in fact it's the default for A256KW. Intuitively I don't see what the key encryption algo has to do with the content encryption algo. I guess some key encryption algos can only handle a certain size key (e.g. a one-time pad). However I don't think that's the case with any of the supported ones...

lestrrat:
Yeah, I hear you. I'm game as long as things work out in the end. Please leave a big fat note somewhere (git commit, code comment, whatever) so I can remember in like 5 years' time :)

imirkin:
Hm, so it looks like some of the *CBC modes indeed don't work. They work with e.g. A128KW but not A128GCMKW or ECDH-ES+A128KW. But those do work with A128GCM content encryption for example. Sigh ... more tests = more failures. However I think this is an improvement. I'll do more digging to figure out what's wrong, but don't have time right now.

imirkin:
OK, so these are the tests that are disabled (not counting ECDH-ES, which is just plain disabled):

			// {jwa.ECDH_ES_A128KW, jwa.A128CBC_HS256},
			// {jwa.ECDH_ES_A256KW, jwa.A256GCM}, fails intermittently
			// {jwa.ECDH_ES_A256KW, jwa.A256CBC_HS512},
			// {jwa.A128GCMKW, jwa.A128CBC_HS256},
			// {jwa.A256GCMKW, jwa.A256CBC_HS512},

The intermittent failure is just weird... but at least we know!

imirkin:
Aha, figured out the problem with the GCMKW ones -- that was just a silly bug in my original impl. Not sure what the deal with the ECDH-ES ones is though, but "it was like that when I got there".

lestrrat:
Just double checking: are you done with this PR for now?

imirkin:
Yes, I think so :) But I'll try to look into some of the other issues that these tests uncovered at a later date.

imirkin:
I have some more fixes (one ECDH-ES fix, and I'm guessing another one in the semi-near future), which will benefit from the interop tests added in this PR. Should I just keep adding stuff here, or should I wait for you to be done reviewing this PR and then open new PR(s)?

imirkin:
OK, this is really it. All the stuff that's "supposed" to work works.

Plain ECDH-ES has been resistant thus far to my attempts to fix it. Not sure where the problem is...

I may do another change to clean up / unify some of the encryption logic which is dispersed throughout. Definitely in a separate PR though.

PR #248, opened 2020-12-16
Fix ECDH-ES support

imirkin:
I don't think the key parameter padding is strictly *required*, but I noticed it was a difference, and I think it makes sense to fix it.


lestrrat:
woot! you rock!

PR #249, opened 2020-12-16
DIRECT support

imirkin:
Not sure if I missed something, this was a lot easier than I anticipated.


imirkin:
Just force-pushed to fix a minor unwanted change to the compact setting of the jose encoding logic.

lestrrat:
LGTM. Thanks!

Issue #250, opened 2020-12-16
Add RFC 8037 (EdDSA) support


imirkin:
https://tools.ietf.org/html/rfc8037

This adds some additional curves to be used with ECDH-ES, as well as a new EdDSA algorithm which can be used for signatures.

imirkin:
Looks like ed25519 is in the Go 1.13+ standard library.

golang.org/x/crypto (which is already used for PBES2 support) has a curve25519, which has a X25519 helper.

X448 does not appear to have support by default. There are a couple of implementations of the curve here:

https://pkg.go.dev/github.com/cloudflare/circl/dh/x448
https://pkg.go.dev/gitlab.com/yawning/x448.git

but X448 and Ed448 would need to be implemented on top of those basic primitives. Unclear if it's worth adding dependencies on them to this project.

lestrrat:
No objections to adding support. 

wrt adding libraries, if it seems like overkill there's always the build tag route to make certain libraries optional.

imirkin:
IMO starting with the 25519 curve should be sufficient, and 448 can be added when the standard libraries gain support. It seems like mainstream libraries are just starting to gain support for it, so I expect Go will get a proper implementation eventually.

imirkin:
@lestrrat interesting problem ... so there's a ed25519.Public/PrivateKey type. But there is no curve25519 public/private key type. The way X25519() works is, essentially, a hard-to-reverse multiplication. So there's a "base" point, let's call it "x", and the private key is your number, 'a', and the public key is someone else's private key 'b' times the base point 'x'. And the idea is that a * (b * x) == b * (a * x), and given "a * x" it's hard to compute "a". They do it all in terms of byte slices.

The jwx library sort of depends on there being dedicated types for these. Should I just define some which alias []byte (as ed25519.*Key already do)? Where should those live? In jwk or in jwx directly or ... ?

lestrrat:
I was fooling around with the same stuff earlier, and I think the best way to go about it is to define `jwk.OctetKey` as `type OctetKey []byte`? I haven't gotten as far as actually making them computations, so I'm not sure if that will suffice, but from the RFC it sounded like it.

Also, you probably need to define `jwa.OctetKeyPair` as `"OKP"`, and `jwa.Ed25519` + `jwa.X25519`. Former for signatures and latter for key agreement.

lestrrat:
Oh, and probably `jwa.EdDSA` as a SignatureAlgorithm

[...]

PR #252, opened 2020-12-18
Add EdDSA signature/verify with Ed25519 curve and ECDH-ES(+*) encryption with X25519 curve support

imirkin:
I haven't tested this with any external applications. jose does not support this. There are internal round-trip tests, but those just validate that we have an equal number of bugs on the encrypt/sign and decrypt/verify paths.

There's at least a node-based jose which does support it, but I haven't gotten around to figuring out how to operate it. Feedback welcome as always.

This addresses the 25519-curve-based part of #250. I think the 448 curve should probably wait until there are more "mainstream" implementations of it (i.e. x/crypto or stdlib).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,6 +27,7 @@ _testmain.go
 .idea
 .vscode
 .DS_Store
+*~
 
 coverage.out
 go.sum
--- a/README.md
+++ b/README.md
@@ -261,12 +261,13 @@ func main() {
 
 Supported key types:
 
-| kty | Curve                   | Go Key Type                            |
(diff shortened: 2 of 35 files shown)


lestrrat:
I think we're okay to go ahead with merging this, but just as a precaution I think a slight clarification in the README that EdDSA and Ed25519 are experimental?

imirkin:
README.md updated with notes on the new bits.

lestrrat:
LGTM. Just waiting for the CI..

lestrrat:
Merged. Thank you !

PR #253, opened 2020-12-18
Reinstate codecov uploads

imirkin:
The tip strategy was removed a while back. Update codecov to upload for the 1.15 go version.

Not 100% sure on how the codecov upload works, so just guessing here... should it somehow be disabled for PRs?

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Test with coverage
         run: make cover
       - name: Upload code coverage to codecov
-        if: matrix.go == 'tip'
+        if: matrix.go == '1.15'
         uses: codecov/codecov-action@v1
         with:
           file: ./coverage.out


lestrrat:
tip was removed because it was doing something incompatible with the rest of the code -- I forget what, but it seemed like a temporary thing, so I meant to reinstate it one of these days.

I have a feeling it might be better to enable codecov on all pull requests and pushes to the master branch

imirkin:
OK, well as evidenced by the codecov report in this PR, it seems to apply to everything that the automated builds apply to.

lestrrat:
Right, on top of that I was actually thinking if we could/should do something like

```
if: ${{ github.event_name == 'pull_request' || github.event_name == 'push' && github.ref == 'refs/heads/master' }}
... run codecov
```

instead of only running for a particular go version.

imirkin:
Won't that cause the codecov to get uploaded twice, once for each version of Go? Why would you want that? And how would the codecov site know the difference -- commit would be the same.

Doing PR + push to master makes sense to me. No clue if that syntax actually works, I'm completely new to the github CI stuff.

lestrrat:
Huh, I had a vague recollection that codecov indexed go versions (or whatever environment variable), but I looked at the codecov page, and apparently not.

Okay, the thing about having to change go versions in that if statement from time to time bothers me ([only if Github would support YAML anchors...](https://github.community/t/support-for-yaml-anchors/16128)), but I think this does the job.

lestrrat:
Thank you!

PR #255, opened 2020-12-18
jwt: remove unused writeJSON function, fix typo in error

imirkin:
Noticed when running staticcheck over the repository.


lestrrat:
This PR is nice, but I wonder why this wasn't caught by golangci-lint. I guess `unused` isn't on by default? hmm.

PR #256, opened 2020-12-18
remove unused json tags, fix other go vet complaints

imirkin:
The only semi-serious one is making the MarshalJSON receiver a pointer
due to the mutex lock.

### DIFF
--- a/jwe/headers_gen.go
+++ b/jwe/headers_gen.go
@@ -72,22 +72,22 @@ type Headers interface {
 }
 
 type stdHeaders struct {
-	agreementPartyUInfo    *buffer.Buffer                  `json:"apu,omitempty"`      //
-	agreementPartyVInfo    *buffer.Buffer                  `json:"apv,omitempty"`      //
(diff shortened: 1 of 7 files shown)


lestrrat:
re: pointer receiver, oops, nice catch.

the tags are just artifacts from the past, so good riddance. yay

imirkin:
FWIW these are the three tools I ran:

go vet
staticcheck (honnef.co/go/tools/cmd/staticcheck)
gosec (github.com/securego/gosec/cmd/gosec)

This is the change for the go vet fixes. The unused code from earlier was found by the staticcheck tool. gosec warns about all sorts of things, but it's mostly not interesting (in generators, or using not-great ciphers).

lestrrat:
Yeah, my assumption was that golangci-lint would do most everything you mentioned. I didn't set the Github Actions + golangci-lint myself, so I guess I got lazy, but I will see if I can do everything in one go with golangci-lint or if I need to add extra calls to the above mentioned tools

PR #261, opened 2020-12-28
Adding jwk set expiration for #260

### LINKED ISSUE #260: Add Expiration field for jwk.Set that is populated by cache-control header
**Abstract**
It would be great if jwk.Set had an `Expiration` property that is populated by the HTTP `cache-control` header when `FetchHTTP` is used or can be set manually if one of the `Parse` methods is used. Additionally, an 'Origin' property to track where the jwk.Set originated from would be helpful (per the usecase below)

Before enumerating available Keys in a Set, one could:
```go
  if set.IsExpired() {
    set, err := set.Refetch()
    if err != nil {
      log.Printf("failed to refetch expired JWKS: %s", err)
      return
    }
  }
```

**Analysis**
Since `FetchHTTP` methods do not make any of the server's headers available to the caller, the only currently alternative is to re-implement `FetchHTTP` outside of jwx.

**Additional context**
Happy to provide a PR if the maintainers agree it would be useful.

shaunco:
fixes #260

### DIFF
--- a/jwk/interface.go
+++ b/jwk/interface.go
@@ -2,6 +2,7 @@ package jwk
 
 import (
 	"crypto/x509"
+	"time"
 
 	"github.com/lestrrat-go/iter/arrayiter"
 	"github.com/lestrrat-go/iter/mapiter"
@@ -41,7 +42,8 @@ const (
 // Set is a convenience struct to allow generating and parsing
 // JWK sets as opposed to single JWKs
 type Set struct {
-	Keys []Key `json:"keys"`
+	Keys    []Key     `json:"keys"`
+	Expires time.Time `json:"expires"`
 }
 
 type HeaderVisitor = iter.MapVisitor
--- a/jwk/jwk.go
+++ b/jwk/jwk.go
@@ -15,7 +15,9 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"strconv"
 	"strings"
+	"time"
 
 	"github.com/lestrrat-go/iter/arrayiter"
 	"github.com/lestrrat-go/jwx/internal/base64"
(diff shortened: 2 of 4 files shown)


imirkin:
Without commenting on implementation, I'd find it useful to have something like this. It had actually not occurred to me that the cache-control header had something useful in it, but I checked and for Azure AD, it's set to 1d, which is fairly reasonable. Currently, I just have an hourly timer which updates the keys, but that timer could be controlled based on the expiration time set in the key set with this API.

Not sure if the IsExpired() bit would be useful to me, but also doesn't hurt anyone. (You don't really want to be fetching an updated keyset while validating a token for a live request...)

jwk/interface.go
```diff
@@ -41,7 +42,8 @@ const (
 // Set is a convenience struct to allow generating and parsing
 // JWK sets as opposed to single JWKs
 type Set struct {
-	Keys []Key `json:"keys"`
+	Keys    []Key     `json:"keys"`
+	Expires time.Time `json:"expires"`
```
imirkin, line 46: RFC7515 makes no mention of this expiration, so probably should not be serialized to/from JSON.

shaunco:
> (You don't really want to be fetching an updated keyset while validating a token for a live request...)

This the the locking that I was referring to in https://github.com/lestrrat-go/jwx/issues/260#issuecomment-751651297 ... and why I left out `Refetch()`. In my use of jwx I use a RWMutex, read for validation, and move to write if expired to refresh, which made sense for me but likely wouldn't make sense in a generic form inside of jwx.

jwk/interface.go
shaunco: As far as I could tell, this `Set` struct is never used for marshal/unmarshal operations, but rather the `proxy` struct at https://github.com/lestrrat-go/jwx/blob/2bc66abc0137afe334b59779962b1fab258c630f/jwk/jwk.go#L291

imirkin: Indeed. So then it should lose its json annotations entirely.

lestrrat:
superseded by #265

PR #262, opened 2020-12-29
Implement jwk.Store types to cache JWKs

lestrrat:


imirkin:
FWIW, I like the impl in #261 better.

 - Doesn't add any new dependencies
 - Lets me supply a default expiration time in case the server doesn't
 - Doesn't (potentially) put an HTTP request in the middle of me getting access to a Set object

lestrrat:
@imirkin The points that kind of get to me about #261 are almost exactly opposite of what you said:

For 1: #261 does hand-rolled interpretation of Cache-Control and Expires header. I would rather have a specific module take care of it
For 3: If the expiration relies on Cache-Control and Expires it inherently /needs/ to go through HTTP requests, I don't want to keep another set of truth (i.e. set.expires) like in #261. 

For 2. This is just not implemented, I was still asking for opinions on #260

lestrrat:
But again, this is still a draft, and I really haven't made up my mind until I see the use-case that I have requested in #260

imirkin:
Expiration is not absolute though. It's max-age, i.e. "you should refetch after this amount of time", not "the content is invalid after this amount of time and must not be used". The keys don't go bad right that second (unlike, say, an access token expiration). It's an indication that it'd be nice to get a fresh set. My current code looks like this:

```
go func() {
	ticker := time.NewTicker(time.Hour)
	defer ticker.Stop()
	for {
		select {
		case <-ticker.C:
			jwks, err = jwk.Fetch()
			// set the jwks on the thing that needs it
		case <-shutdown:
			return
		}
	}
}()
```

And then I just use the jwks in the place that cares without a worry in the world. If the keys go stale, they go stale. No big deal. If the server that provides these goes down (as has happened in the relatively recent past, might I add, with Azure AD), then I don't start failing requests after an hour. The keys change approximately once a year in practice.

Now instead of having an hourly ticker, it'd be nice to make that loop use a `time.Timer` instead, and set it for the expiration time of the request. If instead this gets encompassed in a "Store" object, that's fine, but I want to (a) have up-to-date keys and (b) not have to wait for them when processing a request.

lestrrat:
Okay, so I guess I saw Cache-Control/Expires and thought HTTP and hence JWK caching/validation, when what I had been asked was simply a way to refresh JWKs periodically?

Let me see if I understand the requirements.

Problem:
1. We want to refresh jwks in "decent enough" intervals.
     * @imirkin doesn't need the jwks to be 100% fresh, does @shaunco need them to be fresh? Is it okay to have stale jwks once in a while?
2. These intervals should be either automatically calculated via the source's expiry (Cache-Contro/Expires), or provided explicitly by the user.
     * We use these values as guidance to refresh, not necessarily to be precise about it.
3. Ideally refreshes happen transparently, behind the scenes, and does not introduce waits for the user.

@lestrrat 's peeves:

1. Don't want to bake in too much extra logic/data in the jwks themselves
   * Want to treat jwks as pieces of raw data, don't want to deal with mutability, sharing, etc from within the JWK.
2. Want to make something external to the jwks, so that if somebody requires more fine-tuned custom refresh strategy, it can be developed separately
3. Would rather not handroll Cache-Control parsing. Maybe handroll it, but keep it in a separate module...

@lestrrat's current thinking process

* github.com/gregjones/httpcache can be added by the user easily. Mybe rip this out of the guts.
* Maybe "Store" is too generic. Maybe what the users want is a thing that controls JWK freshness.
    * Maybe AutoRefresh, a type of a jwk.Store?

```go
// * JWKs fetched via this mechanism should be considered read-only
store := jwk.NewAutoRefresh(
  // By default JWKs persist unchanged for 90% of their max-age, at which point
  // it is refreshed. They are also refreshed if they haven't been used for X amount of time
  // (which is when they are evicted from the internal registry)
  //
  // This setting tells the store to ALWAYS use this value for the "max-age" before a refresh.
  // The value is ignored if Cache-Control/Expires values are shorter than the specified amount
  jwk.WithRefreshInterval(15*time.Minute),
  
  // in case you want a custom HTTP client, maybe use httpcache
  jwk.WithHTTPClient(http.Client{ Transport: httpcache.Transport }),
)

// in your main loop...
for {
    // Fetchthe JWK.
    // 
    // The user does not need to check if the keyset is fresh or not.
    //
    // The store object checks the resource pointed by the url needs refreshing,
    // and does this asynchronously.
    keySet, err := store.Fetch(url, 
       // If this is specified, you can control the individual JWK's refresh interval
       jwk.WithRefreshInterval(time.Hour),
       jwk.WithPersist(true), // don't ever evict this entry. default false
    )

    // ... do whatever with the keySet ...

    // perhaps JWKs obtained by this method should be considered read-only.
}
```

shaunco:
Problem:
1. Stale for a bit is fine, but in the case of a key rotation, it is nice to have that happen in somewhere between the 15 seconds Auth0 uses and ~1 day. If the rotation is due to an emergency, we can always restart services to force a rotation.
2. Roughly, yes
3. That would be amazing, but introducing locks is never free. I'm about to start boiling down my use case to some sample code and will include the RWMutex I currently use

Peeves:
all fair, although I would prefer to make use of the hints provided by cache-control.

Thinking:
Can the `Store` or `AutoRefresh` concepts be made in to an interface, where jwx provides an implementations based on httpcache? This allows for @imirkin's model without pulling in httpcache, or the Store model where I can pull in httpcache.

lestrrat:
@shaunco my pseudocode has a `WithHTTPClient` which basically removes the need to have httpcache as a dependency. That leaves the handrolling the Cache-Control parsing bit, whichI haven't thought through

imirkin:
I largely agree with your sentiments, @lestrrat. Not sure about the 90% thing, but also not strongly opposed, I can see arguments for that, and for 100%. Perhaps it would make sense to bind a Store to a particular URL? So that the API becomes like

store := jwk.NewAutoRefresh(url, options...)

And then to use

set := store.Get()

Where the Get() is guaranteed non-blocking? Internally a fetch just updates the latest Set object, so no locking required. (And definitely yes to Set immutability assumption.)

lestrrat:
@imirkin I think that's a non-starter for @shaunco's case, because I think he's writing a gRPC auth interceptor, which may use multiple JWKs within the same interceptor ? I'm guessing, so I may be wrong though.

imirkin:
It still has to know what URL to use, presumably based on some group of parameters (instance, etc). Could have a map of that -> store...

lestrrat:
Gawd, why didn't I make jwk.KeySet an interface. That would have allowed for a middleware type of thingie over the basic jwk.KeySet, like

```go
keyset, _ := jwk.Parse(...)

autorefreshingKeySet := jwk.AutoRefresh(keyset, options...)

autorefreshingKeySet.Refresh()
```

And I think we would've gotten away with the whole lot. Alas. Maybe next major version.


---

But anyway, I guess if @shaunco's use case can tolerate it, a single JWK wrapper/store could definitely be an option.

shaunco:
Yes, we can have multiple JWKS that we run through, in order, until one of them works which is why I check for expiration and refetch (if necessary) prior to calling `ParseString()`, which matches based on `kid`. Going down the atomic pointer route lands back at the thundering herd problem with a bunch of concurrent authorization checks each doing their own http fetch and then updating the set pointer.

> But anyway, I guess if @shaunco's use case can tolerate it, a single JWK wrapper/store could definitely be an option.

The nice thing about a single JWK store _(that tracks URI-->set)_ is that retrievals from the store can be made nearly free until expiration, at which point they all block for one fetch, then go back to being nearly free. If the store has an iterator, I'm still in good shape.

imirkin:
@shaunco Well I was thinking that it'd be the store's job to keep things fresh. So the user just gets the latest pointer, whatever that happens to be. And the store, on its own schedule (ideally influenced by the cache-control header), and only once per set, updates it at the appropriate time. If you have 10000 sets that all expire at the same time then that's a bit annoying, perhaps adding some fuzz to the expiration time would be appropriate.

It sounds like you actually want two separate methods:

1. Get the most-recently-fetched, no matter how old, non-blocking
2. Update for-real (but not if there's an update in progress), then return the latest thing

Where you do the first one, and if you don't find the key you're looking for, do the second one. I guess that makes sense ... at least in my use-case, new keys are put into rotation approximately never, and would be added well in advance of being used to actually sign things.

For the latter case, you'd definitely want some sort of locking to avoid multiple identical requests.

@lestrrat I initially thought my use-case was quite similar to @shaunco's, but seems like maybe not. The (current) FetchHTTP function is all of 25 lines long, so perhaps I'd be better off just reimplementing it directly to grab the headers out of the response.

lestrrat:
Okay,  I need to jet for the day.

Please feel free to continue to discuss possible solutions -- but it's the end of the year, and my responses will be slow for the next few days.

@imirkin and @shaunco (especially @shaunco) if you have the time, please see if you can distill your use case in code. pseudocode is fine, we just need to align our expectations.

If not, that's fine too, we'll be sure to catch up come 2021 :)

I may come back later again while it's still 2020, but I will send you with best wishes for your new year's holidays for now.
Happy new year, and thanks for your continued interest in this project!

shaunco:
Example is at https://github.com/shaunco/jwx-refresh ... note that go.mod points to my #261 fork of jwx.

Happy new year to you as well, and thank you for creating such a fantastic package and maintaining it so diligently!

lestrrat:
superseded by #265

PR #265, opened 2020-12-31
Implement jwk.AutoRefresh to automatically refresh JWKS in an asynchronous manner

lestrrat:
refs #260, #261, #262

### DIFF
--- a/jwk/refresh.go
+++ b/jwk/refresh.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 	"time"
 
+	"github.com/lestrrat-go/backoff"
 	"github.com/lestrrat-go/httpcc"
 	"github.com/pkg/errors"
 )
@@ -36,6 +37,9 @@ type AutoRefresh struct {
 }
 
 type target struct {
+	// The backoff policy to use when fetching the JWKS fails
+	backoff backoff.Policy
+
 	// The HTTP client to use. The user may opt to use a client which is
 	// aware of HTTP caching, or one that goes through a proxy
 	httpcl *http.Client
@@ -72,6 +76,9 @@ type target struct {
 	// The timer for refreshing the keyset. should not be set by anyone
 	// other than the refreshing goroutine
 	timer *time.Timer
+
+	// Semaphore to limit the number of concurrent refreshes in the background
+	sem chan struct{}
 }
 
 type resetTimerReq struct {
@@ -139,8 +146,11 @@ func (af *AutoRefresh) Configure(url string, options ...AutoRefreshOption) {
 	var hasRefreshInterval bool
 	var refreshInterval time.Duration
 	minRefreshInterval := time.Hour
+	bo := backoff.Null()
 	for _, option := range options {
 		switch option.Name() {
+		case optkeyRefreshBackoff:
+			bo = option.Value().(backoff.Policy)
 		case optkeyRefreshInterval:
 			refreshInterval = option.Value().(time.Duration)
 			hasRefreshInterval = true
@@ -181,9 +191,11 @@ func (af *AutoRefresh) Configure(url string, options ...AutoRefreshOption) {
 		}
 	} else {
 		t = &target{
+			backoff:            bo,
 			httpcl:             httpcl,
 			minRefreshInterval: minRefreshInterval,
 			url:                url,
+			sem:                make(chan struct{}, 1),
 			// This is a placeholder timer so we can call Reset() on it later
 			// Make it sufficiently in the future so that we don't have bogus
 			// events firing
@@ -220,7 +232,13 @@ func (af *AutoRefresh) releaseFetching(url string) {
 	af.muFetching.Unlock()
 }
 
-// Fetch retrieves the *jwk.Set object associated with the `url`.
+func (af *AutoRefresh) getRegistered(url string) (*target, bool) {
+	af.muRegistry.RLock()
+	t, ok := af.registry[url]
+	af.muRegistry.RUnlock()
+	return t, ok
+}
+
 // If it has previously been fetched, then a cached value is returned.
 // If this the first time `url` was requested, an HTTP request will be
 // sent, synchronously. Also, when accessed via multiple goroutines
@@ -231,18 +249,31 @@ func (af *AutoRefresh) releaseFetching(url string) {
 // DO NOT modify the *jwk.Set object returned by this method, as the
 // objects are shared among all consumers and the backend goroutine
 func (af *AutoRefresh) Fetch(ctx context.Context, url string) (*Set, error) {
-	af.muRegistry.RLock()
-	_, ok := af.registry[url]
-	if !ok {
+	if _, ok := af.getRegistered(url); !ok {
 		return nil, errors.Errorf(`url %s must be configured using "Configure()" first`, url)
 	}
-	af.muRegistry.RUnlock()
 
 	ks, found := af.getCached(url)
 	if found {
 		return ks, nil
 	}
 
+	return af.refresh(ctx, url)
+}
+
+// Refresh is the same as Fetch(), except that HTTP fetching is done synchronously.
+// This is useful for when you want to force an HTTP fetch instead of waiting
+// for the background goroutine to do it, for example when you want to
+// make sure the AutoRefresh cache is warmed up before starting your main loop
+func (af *AutoRefresh) Refresh(ctx context.Context, url string) (*Set, error) {
+	if _, ok := af.getRegistered(url); !ok {
+		return nil, errors.Errorf(`url %s must be configured using "Configure()" first`, url)
+	}
+
+	return af.refresh(ctx, url)
+}
+
+func (af *AutoRefresh) refresh(ctx context.Context, url string) (*Set, error) {
 	// To avoid a thundering herd, only one goroutine per url may enter into this
 	// initial fetch phase.
 	af.muFetching.Lock()
@@ -265,13 +296,13 @@ func (af *AutoRefresh) Fetch(ctx context.Context, url string) (*Set, error) {
 		defer af.releaseFetching(url)
 
 		// The first time around, we need to fetch the keyset
-		if err := af.refresh(ctx, url); err != nil {
+		if err := af.doRefreshRequest(ctx, url, false); err != nil {
 			return nil, errors.Wrapf(err, `failed to fetch resource pointed by %s`, url)
 		}
 	}
 
 	// the cache should now be populated
-	ks, ok = af.getCached(url)
+	ks, ok := af.getCached(url)
 	if !ok {
 		return nil, errors.New("cache was not populated after explicit refresh")
 	}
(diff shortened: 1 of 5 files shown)


lestrrat:
@imirkin @shaunco

Okay, here's a PoC of an asynchronous JWKS refreshing tool

intended usage:

```go
af := jwk.NewAutoRefresh(ctx)

for { // main app loop
    keyset, err := af.Fetch(ctx, url, ...) // options
    if err != nil { // will most likely only error when the initial fetch fails, as the key is cached
        ...
    }
}
```

imirkin:
I could be missing something, but isn't this missing the cache control thing? Or is that to come later? Looks like this covers my use-case ... gets me a store which I can shut down whenever I want.

The reflect.Select usage is overkill for my case, but like you say in the comments, doesn't actually matter.

lestrrat:
Yes, I skipped the Cache-Control part because, well, I didn't feel like it ATM :)
BTW, I did write a HTTP/1.1 Cache-Control header parser [here](https://github.com/lestrrat-go/httpcc) so I can incorporate it later.

shaunco:
I like this direction. Being able to `Fetch` a set every time, and not needing to think about expired/fresh is nice.

... and happy to pull in your httpcc package for my usecase!

lestrrat:
I'm aware of LoadOrStore, but I always fail to understand how to use it properly.

This doesn't work, does it? subsequent Loads will get a nil, and will just get stuck.

```go
ch, loaded := url2ch.LoadOrStore(url, nil)
if loaded {
   <-ch
} else { .... }
```

And this next one works, but it isn't ideal, because we would be generating a new chan for every read, right?

```go
newCh := make(chan struct{})
ch, loaded := url2ch.LoadOrStore(url, newCh)
if loaded {
   <-ch
} else { .... }
```

I think what I want is a write through function, which I pictured to be something like `LoadOrStore(key, func() interface{})`. The map needs to retain a write lock while it's checking for the existence of a key *and* on a miss, create a new channel + write it to the map, which leads to my code

```go
lock.Lock()
ch, loaded := url2ch[url]
if loaded {
   lock.Unlock()
   <-ch
} else {
   ch := make(chan struct{})
   url2ch[url] = ch
   lock.Unlock()
   ...
  close(ch)
}
```

I'm not quite sure how to achieve this with `sync.Map`

lestrrat:
Also: I /like/ manual locking :-p I get a kick out of synchronizing concurrent code.

lestrrat:
@shaunco @imirkin Added HTTP/1.1 Cache-Control and Expires header handling, which I basically stole from #260

The API is still ambiguous about what happens when you do

```go
af.Fetch(..., jwk.WithMinRefreshInterval( time.Minute ) ) // initial call
// subsequent calls...
af.Fetch(..., jwk.WithMinRefreshInterval( time.Hour ) )  // value changed
af.Fetch(..., jwk.WithRefreshInterval( time.Second) ) // wow, now we're setting it explicitly?
af.Fetch(..., jwk.WithRefreshInterval( ??? ) ) // how do I reset this?
```

lestrrat:
Maybe provide a separate API to configure a URL, and an API to fetch the URL:

```go
ar.Configure(url, jwk.WithMinRefreshInterval( ... ))
```

So the general use case will become

```go
ar := jwk.NewAutoRefresh(ctx)
ar.Configure(...)

for { // main app loop
    keyset, err := ar.Fetch(ctx, url)
    ...
}
```

lestrrat:
Whoof. It's 23:24 on New Years Day, and I finally figured out the error in the test case. It was an error in the test (and I'm glad it wasn't the code)

So at this point I think I nailed what I currently consider to be the more-or-less correct API for this AutoRefresh thing.

The boiler plate code would look like this:

```go
af := jwk.AutoRefresh(ctx)
af.Configure( url1, options ... )
af.Configure( url2, options ...)

for {
   keyset, err := af.Fetch(ctx, url1) // or url2, whatever
   ...
}
```

The configure bit was definitely necessary to avoid confusion in the user-side.
It was really weird when `af.Fetch()` took options, as the only thing we could do
was to either ignore new options from the second call on, or completely overwrite
the previous options unless you called `af.Fetch()` with exactly the same values.
By having a `af.Configure()` method, we're at least making explicit when we change
the settings.

The side effect of this API change is that you MUST pre-register the URL before
using `af.Fetch()`. I don't know if this causes problems for you two. Let me know,

shaunco:
I'm good with pre-registering the URL.

I really appreciate all your thought and effort on this!

jwk/refresh.go
```diff
@@ -0,0 +1,403 @@
...
+}
+
+func calculateRefreshDuration(res *http.Response, refreshInterval *time.Duration, minRefreshInterval time.Duration) time.Duration {
+	// This always has precedence
+	if refreshInterval != nil {
+		return *refreshInterval
+	}
+
+	if v := res.Header.Get(`Cache-Control`); v != "" {
+		dir, err := httpcc.ParseResponse(res.Header.Get(`Cache-Control`))
```
imirkin, line 374: ParseResponse(v)

jwk/refresh.go
```diff
@@ -0,0 +1,403 @@
...
+
+	// To avoid a thundering herd, only one goroutine per url may enter into this
+	// initial fetch phase.
+	af.muFetching.Lock()
+	fetchingCh, fetching := af.fetching[url]
+	// unlock happens in each of the if/else clauses because we need to perform
+	// the channel initialization when there is no channel present
+	if fetching {
+		af.muFetching.Unlock()
+		<-fetchingCh
```
imirkin, line 198: This should also select on ctx.Done(). Otherwise you could end up waiting too long.

jwk/refresh.go
```diff
@@ -0,0 +1,403 @@
...
+		// The first time around, we need to fetch the keyset
+		if err := af.refresh(ctx, url); err != nil {
+			return nil, errors.Wrapf(err, `failed to fetch resource pointed by %s`, url)
+		}
+	}
+
+	// the cache should now be populated
+	ks, ok = af.getCached(url)
+	if !ok {
+		panic("cache was not populated after explicit refresh")
```
imirkin, line 216: That should just be an error. The site could be down. Don't want to start throwing panics in that case. The "main" fetcher will return an error from the refresh() function, but the waiters will panic, I don't think that's good.

jwk/refresh.go
```diff
@@ -0,0 +1,403 @@
...
+			// Do not fire a refresh in case the channel was closed.
+			if !recvOK {
+				continue
+			}
+
+			// Time to refresh a target
+			t := targets[chosen-baseidx]
+
+			//nolint:errcheck
+			go af.refresh(context.Background(), t.url)
```
imirkin, line 310: should this use the context passed into this function instead?

imirkin (commented):
A few things I noticed included here. Overall I like the API. I do wonder if the first fetch should somehow be treated specially (and separately) from the cache lookup, but I also don't have concrete suggestions. Maybe it's fine as-is... I do wonder what happens when it fails -- I guess the next time we try to fetch, it'll try to re-retrieve it, etc etc, and we can't really do much better than that?

lestrrat:
@imirkin Point taken. Let me think about that after I fix the bits you pointed out.

lestrrat:
@imirkin re: failure on Fetch; my take was yes, if the Fetch fails, I figured there's no much else we can do other than keep trying to fetch, and report them errors. The only other thing I thought about was maybe adding a negative cache. Is there any scenario in particular that is worrying you?

lestrrat:
Personal memo. TODOs

* [ ] Documentation
* [ ] Debugging aids -- I want to know all of the JWKs and their next refresh time

imirkin:
> @imirkin re: failure on Fetch; my take was yes, if the Fetch fails, I figured there's no much else we can do other than keep trying to fetch, and report them errors. The only other thing I thought about was maybe adding a negative cache. Is there any scenario in particular that is worrying you?

Right now, I just bail on starting the server if there's an error fetching jwks. And then surrounding logic will try starting it again in a few minutes. Not sure if that's the best thing though -- starting it up without a valid jwks seems OK, but then the internal logic should retry "more often" to get a valid jwks.

An additional thing that I just thought about was what if the *second* fetch fails ... you'd want to retry more aggressively to fetch an updated set, but again not in a tight loop... not sure how that's handled by your current logic.

lestrrat:
What if we have a synchronous `Refresh(ctx, url)` method? That would allow the user to handle warm-ups where things must succeed, and provide a way to fallback when fetches are failing?

Trying to figure out a failed refresh after the refresh started is a bit hard... If this is required, I think we would need an event-dispatch mechanism where the consumer can listen for failed refresh events.

imirkin:
> What if we have a synchronous `Refresh(ctx, url)` method? That would allow the user to handle warm-ups where things must succeed, and provide a way to fallback when fetches are failing?

Works for me. But not 100% sure it's actually required ... I think if we do the below, that may be sufficient:

> 
> Trying to figure out a failed refresh after the refresh started is a bit hard... If this is required, I think we would need an event-dispatch mechanism where the consumer can listen for failed refresh events.

I think just having a refresh interval on error setting would be sufficient -- i.e. tell it if there's a fetch error, just try again in 5 minutes. Or whatever is configured by Configure().

lestrrat:
@imirkin I thought about this some more, and I distilled the problem to the need to  keep fetching the JWKS until it succeeds in a loop, but using some sort of delays in between. Which calls for a backoff algorithm.

So I did just that. You can specify a backoff algorithm to be used in the background refreshing goroutine using `WithRefreshBackoff()`. This forces the background refresh goroutine to go in a loop when the fetch fails. You can make this backoff to, for example, try 10 times while waiting 15 seconds in between each fetch:

```go
af.Configure(url, jwk.WithRefreshBackoffPolicy(
   backoff.Constant(backoff.WithInterval(15*time.Second), backoff.WithMaxRetries(10)),
))
```

Or use an exponential backoff with maximum of 1 hour intervals in between fetches

```go
af.Configure(url, jwk.WithRefreshBackoffPolicy(
   backoff.Exponential(backoff.WithMaxInterval(time.Hour)),
)
```

I hope you don't mind using one more of my libraries for this :) And while "If the HTTP fetch fails, just try again in X amount of time" sounded simple and appealing, I could feel the background refreshing goroutine's `select {}` loop's semantics being overloaded, so opted for a more straight forward "just keep trying until you succeed" approach.

And while I was at it, adding `Refresh()` was a real low hanging fruit, so I just added it anyways.

Let me know what you think.

shaunco:
> I distilled the problem to the need to keep fetching the JWKS until it succeeds in a loop, but using some sort of delays in between. Which calls for a backoff algorithm.

I like this being an option, especially since the library already exists. That said, my logic is the same as @imirkin ... if the first fetch fails my server exits, as there is no point in starting/joining a cluster/listening for connections if I can't authenticate them. The k8s cluster will try to start it again.

imirkin:
@shaunco It's still useful when the fetches fail after the first fetch...

lestrrat:
@shaunco What you want to do, you can do with `Refresh()`, which is a synchronous call to refresh the token

```go
af := jwk.NewAutoRefresh(ctx)
af.Configure(url, jwk.WithRefreshBackoff( ... ) )
// Make sure to fetch it once. This is synchronous, so you can make sure
// that it succeed at least once before starting the loop
_, err := af.Refresh(url)

for {
   // If the next refresh fails in the background, backoff kicks in in the refresh goroutine, and it will
   // keep retrying until the backoff expires. If the backoff expires and still no go, we keep using the
   // old cached value.
   keyset, err := af.Fetch(url)
   ...
}
```

lestrrat:
if my previous comment makes sense, I think we're all sorted? @imirkin @shaunco, please give me your final yay as a reaction to this comment if we're all good, and I'll merge.

lestrrat:
I think I'm going to go ahead and merge. I won't be doing a release for a few more days, so @imirkin please holler if there  are any problems

shaunco:
Thank you for all the effort on this @lestrrat!

PR #271, opened 2021-01-10
Accept jwk.KeyUsageType values in Set() for jwk.KeyUsageKey

lestrrat:


steevehook (approved):
👍 tested, works beautifully. Thanks for real fast response and fix. You the man

lestrrat:
Cool! I won't be making a release immediately, so please use master for the time being

PR #287, opened 2021-01-15
Allow custom signature generation and verification methods

### LINKED ISSUE #286: Ability to choose underlying implementation of jwa.SignatureAlgorithm
**Abstract**
Provide a way to choose the underlying implementation for a `jwa.SignatureAlgorithm` value during `Sign`. My main objective would be to use [cloudflare/circl](https://github.com/cloudflare/circl)'s implementation of Ed25519 instead of `"crypto/ed25519"` from the standard library. This would provide a huge performance improvement when your architecture can benefit from Cloudflare's optimized code.

I only have read parts of the library code and therefore I'm probably not aware of all the underlying consequences, but maybe there would be a way to expose a new API method that accepts a `jws.Signer` instead of an `jwa.SignatureAlgorithm`.

Because the `alg jwa.SignatureAlgorithm` parameter is only passed down and never actively used until calling `sign.New(alg)` in _jws.go_, we could integrate this in a new API while keeping backwards compatibility and  avoiding duplicate code.

An example of the `jwt` package (the same concept would apply to `jws`):

```go
// Current API (only creates a signer with the alg parameter) and
// calls new API
func Sign(t Token, alg jwa.SignatureAlgorithm, key interface{}, options ...Option) ([]byte, error) {
	signer, err := sign.New(alg)
	if err != nil {
		return nil, errors.Wrap(err, `failed to create signer`)
	}

	return SignWithExplicitSigner(t, signer, key, options...)
}

// New API (all the code from the current Sign API copied but passing
// down a signer into the jws package instead of an algorithm type
func SignWithExplicitSigner(t Token, signer jws.Signer, key interface{}, options ...Option) ([]byte, error) {
var hdr jws.Headers
	for _, o := range options {
		switch o.Ident() {
		case identHeaders{}:
			hdr = o.Value().(jws.Headers)
		}
	}

	buf, err := json.Marshal(t)
	if err != nil {
		return nil, errors.Wrap(err, `failed to marshal token`)
	}

	if hdr == nil {
		hdr = jws.NewHeaders()
	}

	if err := hdr.Set(`typ`, `JWT`); err != nil {
		return nil, errors.Wrap(err, `failed to sign payload`)
	}
	sign, err := jws.SignWithExplicitSigner(buf, signer, key, jws.WithHeaders(hdr))
	if err != nil {
		return nil, errors.Wrap(err, `failed to sign payload`)
	}

	return sign, nil
}
```

**Usage**
Then one could implement their own `jws.Signer` outside of this library (with the benefit of not increasing the footprint of this repository and adding (for most people) unnecessary dependencies (like [cloudflare/circl](https://github.com/cloudflare/circl))).

An example would look like the following:

```go
package circlesigner

import (
	"github.com/cloudflare/circl/sign/ed25519"
	"github.com/lestrrat-go/jwx/jwa"
	"github.com/pkg/errors"
)

type EdDSASigner struct {
}

func NewEdDSA() (*EdDSASigner, error) {
	return &EdDSASigner{}, nil
}

func (s EdDSASigner) Algorithm() jwa.SignatureAlgorithm {
	return jwa.EdDSA
}

func (s EdDSASigner) Sign(payload []byte, keyif interface{}) ([]byte, error) {
	switch key := keyif.(type) {
	case ed25519.PrivateKey:
		return ed25519.Sign(key, payload), nil
	default:
		return nil, errors.Errorf(`invalid key type %T`, keyif)
	}
}
```

**Analysis**
I don't see many drawbacks to this approach, as it only changes a few trivial lines of code that shouldn't create any new problems / bugs **from my point of view** - as mentioned before I have only read parts of the library code :)

Pros:
- Backwards compatible API
- Close to zero new code that needs to be maintained
- Flexible enough that various alternative implementations could be used

**Additional context**

Source of image: https://blog.cloudflare.com/introducing-circl/

image

lestrrat:
What:
* Add `jws.RegisterSigner(jwa.SignatureAlgorithm, jws.SignerFactory)` and `jws.RegisterVerifier(jwa.SignatureAlgorithm, jws.VerifierFactory)` to allow using implementations outside of `github.com/lestrrat-go/jwx` for signature generation and verification
* fixes #286

Side effects:
* deprecate `github.com/lestrrat-go/jwx/jws/sign` and `github.com/lestrrat-go/jwx/jws/verify`


lestrrat:
Added usage sample in `examples/jws_example_test.go`

TODO: Possibly remove the `alg` argument from each of the factory methods. I still think making a factory instead of registering a static object is better because who knows, some signer/verifier may need to store per-instance state information, but I think now that we're registering the factory methods this information seems redundant.

lestrrat:
The sample needed updating. Forgot to fix go.mod to point to a local directory

PR #289, opened 2021-01-16
Optimize jwt and jws parsing

johejo:
- Remove unnecessary type conversions
- Optimize when the input is a byte sequence, *bytes.Buffer, *bytes.Reader or *strings.Reader.
- Add benchmark for `jwt.ParseBytes`

Before
```
goos: linux
goarch: amd64
pkg: github.com/lestrrat-go/jwx/jwt
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
BenchmarkJWTParse-16    	  141610	     24251 ns/op	   16352 B/op	      69 allocs/op
PASS
ok  	github.com/lestrrat-go/jwx/jwt	8.545s
```

After
```
goos: linux
goarch: amd64
pkg: github.com/lestrrat-go/jwx/jwt
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
BenchmarkJWTParse-16    	  175860	     15419 ns/op	    7152 B/op	      63 allocs/op
PASS
ok  	github.com/lestrrat-go/jwx/jwt	7.782s
```

In the real world, io.Reader has some similar to byte sequences.
How about the real world use case of jwx?

### DIFF
--- a/jws/jws.go
+++ b/jws/jws.go
@@ -26,8 +26,10 @@ import (
 	"bytes"
 	"context"
 	"io"
+	"io/ioutil"
 	"strings"
 	"unicode"
+	"unicode/utf8"
 
 	"github.com/lestrrat-go/jwx/internal/base64"
 	"github.com/lestrrat-go/jwx/internal/json"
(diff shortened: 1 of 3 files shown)


lestrrat:
Nice. 

Let me fool around with this for a bit, as I have a couple of things I want to consider. My main concern is the addition of new API. Would it be possible to only introduce ParseBytes() as a public API? This also has to do with the fact that, when the time comes, I've been meaning to change the various Parse(io.Reader) that I had elsewhere in this library with Parse([]byte) and ParseReader(io.Reader) -- I wanted to keep this change until I bump the minor version sometime.

lestrrat:
I made #289. Some random observations

* ParseString doesn't need to roll its own loop to find the first character. ParseBytes([]byte(s)) is *slightly* slower, but not significantly so. I think the tradeoff of having less code is better than having more code.
* The switch statement to optimized *bytes.Buffer et al have been refactored into a function
* The test coverage went down, so added more tests
* I also added removal warnings for v1.1.0

lestrrat:
merged. Thank you!

PR #297, opened 2021-01-17
Implement jwk.PublicRawKeyOf(), jwk.PublicKeyOf(), and jwk.PublicSetOf()

### LINKED ISSUE #293: jwks json marshal
I been using this library for testing an OIDC flow (as a client). I been using the jwa, jwt, and jwk packages you have published here to accomplish that. Works great, thank you.

I think it would be useful to be able to marshal the `jwk.Set` struct into an RFC compliant json structure, so that I can serve it directly as a jwks uri for signature verification. However, this struct is missing the json tag for the [`Keys` field](https://github.com/lestrrat-go/jwx/blob/v1.0.8/jwk/interface.go#L44), as a result the `json.Marshal` is producing non-compliant json ("Keys" instead of "keys"). https://tools.ietf.org/html/rfc7517#section-5.1

I am not sure if there already is an option to do this (I was unable to find one) in your extensive set of libraries. I think just adding the tag `json: "keys"` [here](https://github.com/lestrrat-go/jwx/blob/v1.0.8/jwk/interface.go#L44) is enough to satisfy this feature request.

Another handy feature would be to extend the `jwk.Set` with a method that would be able to produce a **PUBLIC KEY** version of your jwks in JSON format (in the case you have only private keys in it). For example:

```
# generate private key
rsaKey, _ := rsa.GenerateKey(rand.Reader, 2048)

# create a JWK from it
key, _ := jwk.New(rsaKey)
key.Set(jwk.KeyIDKey, "gbolo1")
key.Set(jwk.KeyUsageKey, jwk.ForSignature)
key.Set(jwk.AlgorithmKey, jwa.RS256)

# create a JWKS
jwkSet := &jwk.Set{Keys: []jwk.Key{key}}

# produce a jwks JSON with the public portions of my private key
jwksJsonBytes, err := jwkSet.PublicJSON()

...
# now i can serve this in my http handler so that my jwt tokens can get verified
http.HandleFunc("/jwks", func(w http.ResponseWriter, r *http.Request) {
   w.Write(jwksJsonBytes)
}

# this handler will produce json output like as follows:
## NOTICE: the private portion of the key(s) are omitted
{
  "keys": [
    {
      "use": "sig",
      "kty": "RSA",
      "kid": "gbolo1",
      "alg": "RS256",
      "n": "3YKkI...",
      "e": "AQAB"
    }
  ]
}


```

lestrrat:
fixes #293

### DIFF
--- a/cmd/jwx/jwx.go
+++ b/cmd/jwx/jwx.go
@@ -87,8 +87,14 @@ func doJWK() int {
 	}
 
 	// TODO make it flexible
+	firstKey, ok := key.Get(0)
+	if !ok {
+		log.Printf("empty keyset")
+		return 0
+	}
+
 	var pubkey interface{}
-	if err := key.Keys[0].Raw(&pubkey); err != nil {
+	if err := firstKey.Raw(&pubkey); err != nil {
 		log.Printf("%s", err)
 		return 0
 	}
--- a/examples/jwt_example_test.go
+++ b/examples/jwt_example_test.go
@@ -36,7 +36,7 @@ func ExampleJWT_ParseJWKS() {
 		//   Then jwt.Parse() will automatically find the matching key
 
 		var payload []byte
-		var keyset *jwk.Set
+		var keyset jwk.Set
 		{ // Preparation:
 			// For demonstration purposes, we need to do some preparation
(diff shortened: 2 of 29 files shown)


lestrrat:
Wrong merge target. Closing.

PR #319, opened 2021-01-29
Add backwards compatibility check, remove io/fs

lestrrat:
Running `go mod tidy` after the tests now so that we catch inconsistencies in go.mod against go.
There is a slight chance that this may be reintroduced if the go team changes how the go version contained in `go.mod` is treated (but not expecting changes in that front)


lestrrat:
Messed up. wrong target branch.

PR #324, opened 2021-01-30
Play the coverage game

lestrrat:
Also, add Remove to jwk, and locking.


lestrrat:
I need more tests for jwk, but am too tired.

PR #330, opened 2021-02-11
fix: add return error

orisano:
I added a return error to generated code of MarshalJSON.

I have a question. should I propagate of original error?


lestrrat:
Oh, yes, the original error should be propagated! my bad.

orisano:
Can we use %w? should we use errors.Wrapf?

lestrrat:
Please use errors.Wrap for now. If we are going to use %w, we should change everything at once.

orisano:
I fixed it

lestrrat:
I went through #331 b/c the CI needed to be re-enabled, and files needed to be generated. Thanks.

PR #333, opened 2021-02-15
Implement Custom Fields

### LINKED ISSUE #332: I want to get an object in private claim more easily
Sorry my bad english skills 🙇 

**Abstract**

I want to get an object in private claim more easily.

Currently, If I want to retrieve the objects in private claim as I wanted types, need to create a new token type which is implemented `jwt.Token` interface (like openid package) and pass it to `jwt.Parse` functions with `jwt.WithToken` option.
It's very hard to me to implement this interface to get a single object.

Example payload that object in private claim.

```json
{
  "exp": 1577880060,
  "iat": 1577880000,
  ...
  "profile": {
    "name": "codehex",
    "age": 100
  }
}
```

This is code to understand the "Parse" function's default behavior.
I've retrieved "profile" object as `map[string]interface{}` if not used above ways.

https://play.golang.org/p/7FzKGCHhJZ5

I figured it would be useful to be able to write code like this.

```go
type Profile struct {
    Name string
    Age int
}

token := jwt.New(jwt.WithPrivateClaimAs("profile", Profile{}))
t, _ := jwt.ParseString(tokenStr, jwt.WithToken(token))
profile, _ := t.Get("profile")

fmt.Printf("%#v\n", profile) // main.Profile{Name:"codehex", Age:100}
```

jwt package

```go
func New(options ...TokenOption) Token

func WithPrivateClaimAs(name string, v interface{}) TokenOption
```

**Analysis**

- Create a new option to `jwt.Parse` function.
  - I had no idea how to handle the case where multiple objects are included in a private claim.

lestrrat:
fixes #332

Each message in the `jwe`, `jwk`, `jws`, `jwt` protocols have predefined fields. Any other field which is not defined in the spec can be populated by the user. Howeversince we don't know the type information before hand we could not decode these values into Go objects.

This PR fixes this by creating a global registry of key names to their types. For example, this makes it so that when you either use `jwt.Parse` or `json.Unmarshal`, the value pointed by x-foo-bar is marshaled into `myawesomepkg.Object`.

```go
func init() {
  jwt.RegisterCustomField(`x-foo-bar`, myawesomepkg.Object{})
}
```

This has global effects.

It is the caller's responsibility to implement the proper `json.Unmarshaler` methods for custom unmarshaling.

### DIFF
--- a/internal/json/registry.go
+++ b/internal/json/registry.go
@@ -1,0 +1,53 @@
+package json
+
+import (
+	"reflect"
+	"sync"
+
+	"github.com/pkg/errors"
+)
+
+type Registry struct {
+	mu   *sync.RWMutex
+	data map[string]reflect.Type
+}
+
+func NewRegistry() *Registry {
+	return &Registry{
+		mu:   &sync.RWMutex{},
+		data: make(map[string]reflect.Type),
+	}
+}
+
+func (r *Registry) Register(name string, object interface{}) {
+	if object == nil {
+		r.mu.Lock()
+		defer r.mu.Unlock()
+		delete(r.data, name)
+		return
+	}
+
+	typ := reflect.TypeOf(object)
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	r.data[name] = typ
+}
+
+func (r *Registry) Decode(dec *Decoder, name string) (interface{}, error) {
+	r.mu.RLock()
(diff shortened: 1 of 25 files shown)


Code-Hex:
I've confirmed that it works 🙏 
https://github.com/lestrrat-go/jwx/issues/332#issuecomment-778984721

Issue #334, opened 2021-02-16
Introduce support for ES256K with secp256k1 curve for signing \ veryfying JWT tokens.


eshepelyuk:
**Abstract**  \ **Additional context**

As a user of [Open Policy Agent](https://www.openpolicyagent.org/) I want to be able to use ES256K with secp256k1 to sign \ verify JWT tokens in REGO policies. 
Open Policy Agent uses `jwx` as a backing technology, so the question is addressed here.

lestrrat:
I'm not familiar with with this algorithm, so... PRs welcome

eshepelyuk:
Hello

I'm not familar with Golang and Cryptography at all, so PR from my side will not be delivered.
If you're willing to keep this issue open - then let more experienced guys to implement it it some future, maybe.
Otherwise, it could be closed.

Thnx

imirkin:
It seems like the canonical secp256k1 implementation is in go-ethereum.

https://pkg.go.dev/github.com/ethereum/go-ethereum/crypto/secp256k1

@lestrrat if you think it's reasonable to depend on this package, I can give this a go.

lestrrat:
@imirkin i was actually looking at https://github.com/decred/dcrd/tree/master/dcrec/secp256k1

from what I could tell if we use ethereum-go or btcd 's implementation you would have to import the entire app, which is... meh

anyway that's all i had time to research. will check back tomorrow morning

imirkin:
@lestrrat Hm, yeah, there's no go.mod file there. Annoying. It wraps libsecp256k1 though, which hopefully gives the implementation some credibility.

My more general concern is with depending on encryption from "weird" places. Thus far, we depend on core lib and x/crypto, which is as non-weird as it gets. I've never heard of decred -- perhaps it's a huge project which gets security audits and whatnot, but that should be confirmed. I've heard of ethereum, and the fact that it's just a libsecp256k1 wrapper makes me a bit more comfortable. There are a few other places that wrap libsecp256k1, and it's not a ton of code and should be importable into jwx directly if need be (looks like a BSD license on the core lib).

I believe the Go maintainers' position is that secp256k1 is an esoteric curve only used for blockchains, so it's unlikely to make it to the main libraries. (Similar to x448.)

lestrrat:
my initial reaction is that i would rather not use a C implementation out of concerns from a support perspective.

that library i linked up there seems to be pure Go, which is why was reading up on it

[...]

PR #336, opened 2021-02-17
Add ES256K signature / verification

imirkin:
Fixes issue #334 

Let me know if I should try to make this do optional build for including the decred repo.


lestrrat:
Awesome!

From the Github Actions log, it seems like the build time didn't suffer too much, eh?

That leaves us with the verification that this is actually interoperable, but I guess there's not much else we can do at the moment. "Field testing" is what we need, I guess :)

I'm game. Will merge.

PR #337, opened 2021-02-18
Add interop tests for jws

imirkin:
In the process, it turned out that we were doing PSS signatures wrong. I don't have a spec reference, but this change makes both jose and jwt.io happy.


lestrrat:
Thanks! much appreciated.

PR #340, opened 2021-02-19
cmd/jwx: fix readme typos

imirkin:
I can't actually run the tool due to the go 1.16 requirement, but pretty sure these are right.


lestrrat:
Thanks, I will s/1.16/1.15/ after merging this

PR #346, opened 2021-02-25
Toggle ES256K using a build tag

lestrrat:
Per #345 by default we're going to skip compiling in ES256K.

We WILL NOT be using `database/sql` style explicit importing scheme for the time being. If we do, this will be done in the next API breaking release. Instead we will just toggle these "extra" features using build tags like `jwx_goccy`. For this one we will be using `jwx_es256k`

### DIFF
--- a/jwa/elliptic_gen.go
+++ b/jwa/elliptic_gen.go
@@ -4,6 +4,8 @@ package jwa
 
 import (
 	"fmt"
+	"sort"
+	"sync"
 
 	"github.com/pkg/errors"
 )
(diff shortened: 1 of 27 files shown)


jwa/elliptic_gen.go
```diff
@@ -19,7 +19,6 @@ const (
 	P256                 EllipticCurveAlgorithm = "P-256"
 	P384                 EllipticCurveAlgorithm = "P-384"
 	P521                 EllipticCurveAlgorithm = "P-521"
-	Secp256k1            EllipticCurveAlgorithm = "secp256k1"
```
imirkin, line 22: Why should jwa care? It's a valid constant irrespective of availability of implementation in the jwk/jws libraries presumably, no?

lestrrat: It was inspired from here https://github.com/lestrrat-go/jwx/blob/a7162536a2a50e66cc69bab0e4deb403e3f1fdc0/cmd/jwx/jwa.go#L69

I didn't want secp256k1 to be listed.

But even before that, I had the premonition that somebody was going to ask (in my best Goofy voice) "Why can't I use ES256K when jwa.Secpt256k1 is available???" So I took the opportunity.

imirkin: OK, fair enough.

lestrrat: The amount of boilerplate needed to compile Secp256k1/ES256k out is getting pretty annoying, we should come up with a better interface for... jwx v1.2 or something in the future

PR #351, opened 2021-03-09
jwt: allow setting a custom value to "typ" JOSE header parameter

debug-ito:
`jwt.Sign` function automatically sets `"JWT"` to `typ` JOSE header parameter. Even though the user provides a custom value for `typ` using `WithHeaders` option, it's overwritten by `"JWT"`.

This pull-request changes the above behavior. Now `jwt.Sign` sets `"JWT"` to `typ` only if the user doesn't provide a custom value for it.

Tests are added to jwt_test.go

### DIFF
--- a/jwt/jwt.go
+++ b/jwt/jwt.go
@@ -227,9 +227,12 @@ func Sign(t Token, alg jwa.SignatureAlgorithm, key interface{}, options ...Optio
 		hdr = jws.NewHeaders()
 	}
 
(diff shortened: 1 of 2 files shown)


debug-ito:
I also revised the error message that `jwt.Sign` returns when it fails to set `typ` field.

lestrrat:
Sounds like a plan. Thanks!

PR #360, opened 2021-03-30
jwk: "invalid memory address or nil pointer dereference" error when create new jwk.Key from empty RSA private  key

tgunsch:
Hi,
we upgraded to 1.6.2 and now getting an panic error in our unit tests while creating a new `jwk.Key` from an empty rsa private or public key:
```
Test Panicked
    runtime error: invalid memory address or nil pointer dereference
    /usr/local/go/src/runtime/panic.go:212

    Full Stack Trace
    math/big.(*Int).Bytes(...)
    github.com/lestrrat-go/jwx/jwk.(*rsaPublicKey).FromRaw(0xc0002fed00, 0xc0003ec1d0, 0x0, 0x0)
    github.com/lestrrat-go/jwx/jwk.New(0xa6fd60, 0xc0003ec070, 0x0, 0x0, 0xc0000f6fb0, 0x1)
....
```
So I added a validatation that at least an error is thrown.


lestrrat:
I ended up redoing most the checks, but big thanks for pointing out the edge case. At the end both RSA/ECDSA keys could have caused the same sort of panic, and RSA even more so, because it had more `*big.Int` numbers to deal with.

Thank you!

lestrrat:
@tgunsch also, I tagged v1.1.7-rc1 including this change so it's slightly easier to specify in your go.mod

PR #366, opened 2021-04-04
docs: align package name in ParseRequest examples

loozhengyuan:
This PR changes the package in the examples for `ParseRequest` from `jwthttp` to `jwt`.

Not sure if using `jwthttp` is intended or not but I think that it will be clearer to users if we change this since they belong in the `jwt` package.


lestrrat:
Oh, that looks like a total find-and-replace gone wrong. Thanks.

PR #372, opened 2021-04-16
another typo...only because I love this package.

stensonb:
super simple to use, excellent piece of go here.  thanks!


lestrrat:
lol, thanks!

PR #381, opened 2021-04-29
[jwt] Use key's algorithm instead of headers when jwt.WithKeySet() is used

lestrrat:


blankdots:
the `alg` key is also `OPTIONAL` in the spec https://datatracker.ietf.org/doc/html/rfc7517#section-4.4 
i get `failed to find matching key for verification: invalid signatre algorithm : invalid jwa.SignatureAlgorithm value` when using `WithKeySet` if `alg` key is missing

PR #401, opened 2021-06-26
Fix documentation and examples for OpenID claims

jandd:
This PR replaces examples and documentation strings using jwt.WithOpenIDClaims() with jwt.WithToken(openid.New().


lestrrat:
Thanks!

Issue #402, opened 2021-06-26
Decryption fails if a custom field parser is registered


anatol:
It is a follow-up for https://github.com/anatol/clevis.go/issues/4

I am using github.com/lestrrat-go/jwx v1.2.1 and following code example shows that if a custom field parser is registered then it makes decryption fail:

```go
package main

import (
	"encoding/json"
	"fmt"
	"github.com/lestrrat-go/jwx/jwa"
	"github.com/lestrrat-go/jwx/jwe"
)

// Pin represents the structured clevis data which can be used to decrypt the jwe message
type Pin struct {
	Pin     string      `json:"pin"`
	Tang    *TangPin    `json:"tang,omitempty"`
	Tpm2    *Tpm2Pin    `json:"tpm2,omitempty"`
	Sss     *SssPin     `json:"sss,omitempty"`
	Yubikey *YubikeyPin `json:"yubikey,omitempty"`
}

type TangPin struct {
	Advertisement *json.RawMessage `json:"adv,omitempty"`
	URL           string           `json:"url"`
}

type Tpm2Pin struct {
	Hash    string `json:"hash,omitempty"`
	Key     string `json:"key,omitempty"`
	JwkPub  string `json:"jwk_pub,omitempty"`
	JwkPriv string `json:"jwk_priv,omitempty"`
	PcrBank string `json:"pcr_bank,omitempty"`
	PcrIds  string `json:"pcr_ids,omitempty"`
}

type SssPin struct {
	Prime     string   `json:"p"`
	Threshold int      `json:"t"`
	Jwe       []string `json:"jwe"`
}

type YubikeyPin struct {
	Type      string     `json:"type"`
	Challenge string     `json:"chalelenge"`
	Slot      int        `json:"slot"`
	Kdf       YubikeyKdf `json:"kdf"`
}

type YubikeyKdf struct {
	Type       string `json:"type"`
	Hash       string `json:"hash"`
	Iterations int    `json:"iter"`
	Salt       string `json:"salt"`
}

func main() {
	key := []byte{195, 170, 42, 171, 98, 176, 98, 162, 57, 170, 62, 69, 175, 209, 200, 151, 81, 135, 63, 43, 93, 20, 16, 111, 13, 26, 138, 188, 15, 19, 26, 242}
[...]

PR #410, opened 2021-07-12
Various cleanups including renaming Do() to PostParse()

lestrrat:


lestrrat:
@lack I fiddled around with for a while, and I changed `(jwe.PostParser).Do` to `(jwe.PostParser).PostParse` as you initially suggested. I'm going to be cutting a release soon

lack:
> @lack I fiddled around with for a while, and I changed `(jwe.PostParser).Do` to `(jwe.PostParser).PostParse` as you initially suggested. I'm going to be cutting a release soon

@lestrrat Nice :)

Did you see the other suggestions I made for the API name (comment after #402 closed)?
- `jwe.WithParseFinalizer(...)`
- `jwe.WithKeySetup(...)`
- `jwe.WithKeyFinalizer(...)`

But `PostParser` is good too.  Someone unfamiliar with its purpose may misread it though, as "this function parses a post" instead of "this function runs after the parser" thanks to the incredible vagueness of the English language :)

If you were thinking of changing it, I would vote for `KeyFinalizer`, since that's what I'm using it to do.

lestrrat:
Yes, I did see them, but I ended up liking PostParse() the best. Thanks for the suggestions!

PR #416, opened 2021-07-15
Allows 0 to be valid token expiration instead of -62135596800

code-merc:
If you have a `0` as a token expiration, issued at, or not before claim, validation will fail.  Currently you have to set it to be -62135596800 to pass the `time.IsZero()` check, this pull would allow unix time 0 to pass.

### DIFF
--- a/jwt/validate.go
+++ b/jwt/validate.go
@@ -151,7 +151,7 @@ func Validate(t Token, options ...ValidateOption) error {
 	}
 
 	// check for exp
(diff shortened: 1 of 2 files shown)


lestrrat:
@code-merc Thanks, can you please provide a test case for this?

code-merc:
Added test cases for zero times for unix and go

lestrrat:
Thanks!

PR #421, opened 2021-07-29
Implement ErrorSink method

lestrrat:
closes #420

### DIFF
--- a/Changes
+++ b/Changes
@@ -6,6 +6,9 @@ v1.2.5 (UNRELEASED)
   * Implement RFC7797. The value of the header field `b64` changes
     how the payload is treated in JWS
   * Implement detached payloads for JWS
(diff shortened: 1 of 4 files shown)


rockspore:
This seems to be working well. Thanks!

A question that may not be closely related: if the URL is misconfigured at the first place, it seems the refresh loop will never happen asynchronously even after a `Fetch()` is invoked? The snapshot indicates it stays at:
```
last refreshed at 0001-01-01 00:00:00 +0000 UTC, next refresh at 0001-01-01 00:00:00 +0000 UTC
```
Is this desired?

lestrrat:
@rockspore Thanks for checking. I just did a final renaming of the method to `ErrorSink`, and the error wrapper to `AutoRefreshError`, and am going to merge. I'm going to leave this in master and not release a new version for a few days. Let me know if you find any problems.

PR #432, opened 2021-08-16
Fix typo

arpanetus:
The very small typo. Just thought why not to this.


lestrrat:
Ah, that's a generated file, so you can't fix it there. It's in jwt/internal/cmd/gentoken/main.go if you want to fix it

PR #433, opened 2021-08-16
Fix typo

arpanetus:
becase => because


arpanetus:
@lestrrat  I guess it works this time

lestrrat:
Thanks. Your fix is correct, but you need to commit the result of running the code generation tool.
Run `go generate`, and commit the result, and the tests should pass.

lestrrat:
Well, I woke up, so I'm just going to go ahead and do the last step, and merge your changes!

lestrrat:
fixed via #434

Thanks for the PR!

PR #460, opened 2021-09-05
Add some auto-assign action

lestrrat:


lestrrat:
I dunno if this works, but I'm going to try

PR #463, opened 2021-09-14
chore: add example with import jwk

caiowWillian:
## Changes

added example to signed a JWT with a imported jwk

## motivation

I had to spend a little time figuring out how to import a jwk. I believe the example can help other people who need to do the same. :)


lestrrat:
Thanks!

lestrrat:
@caiowWillian Just out of curiosity, did you look into https://github.com/lestrrat-go/jwx/blob/main/docs/01-jwt.md as well?
If yes, I might also add some docs  about how to use JWKs over there as well

Issue #395, opened 2021-06-14
Cannot verify Azure oauth2 token signature of a valid token.


konradmalik:
**Describe the bug**
Cannot verify Azure oauth2 token signature, while token is definitely good and valid - allows me to access Microsoft Graph.

The error that shows up while I want to validate the token:
```
failed to parse payload: failed to verify jws signature: failed to verify message: crypto/rsa: verification error
```

Pretty sure I must be doing something wrong, but I cannot identify what it is.

go version:
```
go version go1.16.4 darwin/amd64
```

**To Reproduce / Expected behavior**
I'm attaching a full code that uses an existing app registration, assigns a device code, created a cached refresh token, and requests a new token:

main.go
```go
package main

import (
	"context"
	"fmt"
	"io"
	"log"
	"net/http"
	"strings"
	"time"

	"github.com/AzureAD/microsoft-authentication-library-for-go/apps/public"
	"github.com/lestrrat-go/jwx/jwk"
	"github.com/lestrrat-go/jwx/jwt"
)

var cacheAccessor = &TokenCache{"serialized_cache.json"}

func acquireTokenDeviceCode() string {
	config := CreateConfig("config.json")

	app, err := public.New(config.ClientID, public.WithCache(cacheAccessor), public.WithAuthority(config.Authority))
	if err != nil {
		panic(err)
	}

	// look in the cache to see if the account to use has been cached
	var userAccount public.Account
	accounts := app.Accounts()
	for _, account := range accounts {
		if strings.EqualFold(account.PreferredUsername, config.Username) {
			userAccount = account
		}
	}
	// found a cached account, now see if an applicable token has been cached
	// NOTE: this API conflates error states, i.e. err is non-nil if an applicable token isn't
	//       cached or if something goes wrong (making the HTTP request, unmarshalling, etc).
	authResult, err := app.AcquireTokenSilent(context.Background(), config.Scopes, public.WithSilentAccount(userAccount))
	if err != nil {
		// either there was no cached account/token or the call to AcquireTokenSilent() failed
		// make a new request to AAD
		ctx, cancel := context.WithTimeout(context.Background(), 100*time.Second)
		defer cancel()
		devCode, err := app.AcquireTokenByDeviceCode(ctx, config.Scopes)
		if err != nil {
			panic(err)
		}
		fmt.Printf("Device Code is: %s\n", devCode.Result.Message)
		result, err := devCode.AuthenticationResult(ctx)
		if err != nil {
[...]

PR #465, opened 2021-09-15
Add FAQ for Microsoft

johejo:
I think this issue is worth adding to the FAQ.
You can fix title or add any cool additional messages.

related #395

### DIFF
--- a/docs/99-faq.md
+++ b/docs/99-faq.md
@@ -73,3 +73,7 @@ Often times we have people asking us about github.com/lestrrat-go/jwx/jwt not be
 For example, when a provider says they will give you an "access token" ... well, it *may* be a JWT, but often times they are just some sort of string key (which will definitely parse if you pass it to `jwt.Parse`). Sometimes what you really want is stored in a different token, and it may be called an "ID token". Who knows, these things vary between implementation to implemention.
 
 After all, the only thing we can say is that you should check that you are parsing. 
+
+## I can't verify ID token from Microsoft with key sets provided by Microsoft jwks uri.
+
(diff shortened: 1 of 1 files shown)


johejo:
BTW, node-jsonwebtoken detects key's algorithm with this hack.
https://github.com/auth0/node-jsonwebtoken/blob/74d5719bd03993fcf71e3b176621f133eb6138c0/verify.js#L113-L118

lestrrat:
I don't get how https://github.com/lestrrat-go/jwx/pull/465#issuecomment-920005781 has to do with #395?

lestrrat:
Oh, the initial question was about JWKs, not IDToken vs AccessToken. I get it.
But no, no automatic algorithm choosing in this library. I would rather err on the side of safety

johejo:
Don't worry about BTW comments.

> But no, no automatic algorithm choosing in this library. I would rather err on the side of safety

I agree.

I wondered why other libraries could validate Azure ID tokens with jwk without alg so easy and I suspected a vulnerability.
But it was magic for specs (and implementations) not vulnerabilities.

lestrrat:
I suppose I _could_ add an option with a big warning:

```
jwt.Parse( ...., jwt.ThisOptionCouldMakeYourAppVulnerable(true))
```

I guess as long as the user can be made aware that they are risking something for ease of use, it could be added

johejo:
Guessing based on kty may be somewhat safe.

Hmm, I'm just hoping that Microsoft will add alg.

lestrrat:
I had an epiphany wrt to the `alg` situation. Please review if you are interested https://github.com/lestrrat-go/jwx/tree/infer-algorithm

PR #466, opened 2021-09-15
Add options to validate claim with function

johejo:
I'm looking for a more appropriate option name.

### DIFF
--- a/jwt/validate.go
+++ b/jwt/validate.go
@@ -53,6 +53,9 @@ func Validate(t Token, options ...ValidateOption) error {
 	var deltas []delta
 	requiredMap := make(map[string]struct{})
 	claimValues := make(map[string]interface{})
+	claimFuncs := make(map[string]claimFn)
+	claimStringFuncs := make(map[string]claimStringFn)
+	claimFloat64Funcs := make(map[string]claimFloat64Fn)
 	for _, o := range options {
 		//nolint:forcetypeassert
 		switch o.Ident() {
@@ -89,6 +92,15 @@ func Validate(t Token, options ...ValidateOption) error {
 		case identClaim{}:
 			claim := o.Value().(claimValue)
 			claimValues[claim.name] = claim.value
+		case identValidateClaimFn{}:
+			pair := o.Value().(claimFnPair)
+			claimFuncs[pair.Name] = pair.Fn
+		case identValidateClaimStringFn{}:
+			pair := o.Value().(claimStringFnPair)
+			claimStringFuncs[pair.Name] = pair.Fn
+		case identValidateClaimFloat64Fn{}:
+			pair := o.Value().(claimFloat64FnPair)
+			claimFloat64Funcs[pair.Name] = pair.Fn
 		}
 	}
 
@@ -181,6 +193,56 @@ func Validate(t Token, options ...ValidateOption) error {
 	for name, expectedValue := range claimValues {
 		if v, ok := t.Get(name); !ok || v != expectedValue {
 			return fmt.Errorf(`%v not satisfied`, name)
+		}
+	}
+
+	for name, fn := range claimFuncs {
+		tv, ok := t.Get(name)
+		if !ok {
+			return fmt.Errorf("%v not satisfied", name)
+		}
+		ok, err := fn(tv)
+		if err != nil {
+			return err
+		}
+		if !ok {
+			return fmt.Errorf("%v not satisfied", name)
+		}
+	}
+
+	for name, fn := range claimStringFuncs {
+		tv, ok := t.Get(name)
+		if !ok {
+			return fmt.Errorf("%v not satisfied", name)
+		}
+		tvv, ok := tv.(string)
+		if !ok {
+			return fmt.Errorf("invalid type for %v, want string but got %T", name, tv)
+		}
+		ok, err := fn(tvv)
+		if err != nil {
+			return err
+		}
+		if !ok {
+			return fmt.Errorf("%v not satisfied", name)
+		}
+	}
+
+	for name, fn := range claimFloat64Funcs {
+		tv, ok := t.Get(name)
+		if !ok {
+			return fmt.Errorf("%v not satisfied", name)
+		}
+		tvv, ok := tv.(float64)
+		if !ok {
+			return fmt.Errorf("invalid type for %v, want float64 but got %T", name, tv)
+		}
+		ok, err := fn(tvv)
+		if err != nil {
+			return err
+		}
+		if !ok {
+			return fmt.Errorf("%v not satisfied", name)
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


lestrrat:
Haven't thought about the name yet, but please do this:

* Please make the primary validator an interface, not a raw function (e.g. `type ClaimValidator interface { Validate(interface{}) error }`)
* Please provide a function type for convenience (e.g. `type ClaimValidatorFunc func(interface{}) error`)

The above is a MUST for my libraries. I don't allow raw `func`s to be passed as arguments.

Below is up to consideration:

* Why should a validating function return _both_ `bool` and `error`? I think a single return value of `error` is enough.
* What happens if claims are inter-dependent? e.g. the correct value for claim A  might change depending on the value of another claim B. Would it be better if this is a function to validate the entire token, and not single claims?

jwt/validate_test.go
```diff
@@ -256,3 +257,63 @@ func TestGHIssue10(t *testing.T) {
 		}
 	})
 }
+
+type validateFn func(interface{}) error
```
lestrrat, line 261: This should be an exported type. Then the consumers of the API can just convert types on the fly like you do in the test
```
type ClaimValidatorFunc func(interface{}) error
```

jwt/validate.go
```diff
@@ -89,6 +90,9 @@ func Validate(t Token, options ...ValidateOption) error {
 		case identClaim{}:
 			claim := o.Value().(claimValue)
 			claimValues[claim.name] = claim.value
+		case identClaimValidator{}:
+			pair := o.Value().(claimValidatorPair)
+			claimValidators[pair.Name] = pair.Validator
```
lestrrat, line 95: Are there cases where the user might want to place multiple validators on a claim? If so this may need to be []ClaimValidator

jwt/validate_test.go
```diff
@@ -256,3 +257,63 @@ func TestGHIssue10(t *testing.T) {
...
+		t.Run("object is expected", func(t *testing.T) {
+			t.Parallel()
+			t1 := jwt.New()
+			assert.NoError(t, t1.Set("object", map[string]interface{}{"k": "v"}))
+			assert.NoError(t, jwt.Validate(t1, jwt.WithClaimValidator("object", fn)))
+		})
+		t.Run("object does not exist", func(t *testing.T) {
+			t.Parallel()
+			t1 := jwt.New()
+			assert.NoError(t, t1.Set("oobject", map[string]interface{}{"k": "v"}))
```
lestrrat, line 297: nit: please make "wrong" values obviously wrong. i.e. it is hard to tell if "oobject" is intentional or just a typo. better to use something like "non-existent-key"

jwt/validate_test.go
```diff
@@ -256,3 +257,63 @@ func TestGHIssue10(t *testing.T) {
...
+			}
+			if v != "v" {
+				return err2
+			}
+			return nil
+		})
+		t.Run("object is expected", func(t *testing.T) {
+			t.Parallel()
+			t1 := jwt.New()
+			assert.NoError(t, t1.Set("object", map[string]interface{}{"k": "v"}))
```
lestrrat, line 291: nit: it may not happen now, but I have been trained to avoid linter warnings ... and I think using the string literal "object" over and over may cause linters to start yelling at me. Can you make this a const, like `const testClaimKey = "object"` and use that throughout

lestrrat (commented):
Other than my comments, I'm still not 100% sure that creating a validator for a single claim is good enough. I mean, what if somebody comes up with a requirement that claim A and claim B have the same value. How would they validate this using this mechanism.

I'm thinking that if we're going to allow users to specify their validation code to be executed inside the Validate function, the main interface should accept the entire token -- but we can create a pre-defined validator that can check for a single claim. What do you think?

```go
type ClaimsValidator interface {
   Validate(Token) error
}

type SingleClaimValidator interface { // name TBD
  Validate(interface{}) error
}

type SingleClaimValidatorFunc func(interface{}) error

type singleClaimValidator struct {
  name string
  claimValidator SingleClaimValidator
}

func (scv *singleClaimValidator) Validate(tok Token) error {
  v, ok := tok.Get(scv.name)
  if !ok {
    return nil
  }
  return scv.claimValidator.Validate(v)
}

var validator = jwt.NewSingleClaimValidator(name, jwt.SingleClaimValidatorFunc(func(v interface{}) error {
  ....
}))
jwt.Validate(tok, jwt.WithClaimValidator(jwt.SingleClaimValidator(validator)))
```

jwt/validate.go
```diff
@@ -184,5 +187,51 @@ func Validate(t Token, options ...ValidateOption) error {
...
+type SingleClaimValidator interface {
+	Validate(interface{}) error
+}
+
+// SingleClaimValidatorFunc is a named type to wrap raw function as SingleClaimValidator.
+type SingleClaimValidatorFunc func(interface{}) error
+
+func (f SingleClaimValidatorFunc) Validate(c interface{}) error {
+	return f(c)
+}
```
johejo, line 214: SingleStringClaimValidator, SingleFloat64ClaimValidator, it would be convenient to have those.

lestrrat:
Alright, let's now think about names :)

There are several ways we can name things. I honestly don't know which one is better yet. I'm just throwing out ideas

I think `jwt.ClaimsValidator` is fine as it is. It's short enough, it's clear what it does.

One idea about `SingleClaimValidator` is, maybe make it `jwt.ClaimValidator`. But it's confusing to the eye, since it only has one character difference with `jwt.ClaimsValidator`. 
For convenience functions, maybe create a separate package? `claims` under `jwt`?

```go

import (
  ...
  "github.com/lestrrat-go/jwx/jwt/claims"
)

jwt.Parse(..., jwt.WithClaimsValidator(claims.ValidateFloat64(`claimName`))
```

Hmmm, maybe also move `jwt.SingleClaimValidator` to this package? `claims.ClaimValidator` ? meh, I don't know if that names is good. but moving packages is a possibility, as it allows more concise, specialized names.

johejo:
I generally prefer a single jwt package.
The `SingleClaimValidator` contains 3 words, I don't think this is too confusing.
I feel that 4 or more words (e.g. SingleStringClaimValidator) are certainly too long, but IMO it's not so painful.

I think `jwt/claims` a well-named subpackage.

However, need to think carefully about whether to place the some types in the `jwt` or `jwt/claims`.

lestrrat:
I would define `jwt/claims` as follows, and I think it makes sense.

```
Package claims contains utilities to that helps users deal with interacting and validating the claims in a `jwt.Token`
```

lestrrat:
I have a few ideas. will get back once I try some stuff out

lestrrat:
Okay, I tried various different implementations, and I just couldn't come up with a good "single claim validation" API.
I'm willing to merge what I have in https://github.com/lestrrat-go/jwx/tree/johejo-add_options_to_validate_claim_with_func.
It may not be exactly what you wanted, but it at least laves the possibility of implementing something like a SingleClaimValidator in the future (because the API doesn't have anything like it), and is still generic enough that you can do anything with it, just not as part of this library.

lestrrat:
#477 is now ready. I gutted `jwt.Validate()` to completely use the `jwt.Validator` interface.
@johejo Let me know what you think, and if we can merge this we can probably do a release soon after

PR #468, opened 2021-09-17
Optimize iteration

johejo:
Iteration of lestrrat-go/iter using goroutine and channel is sometimes useful with a generic interface, but it can be slow.

I tried the benchmark for `LookupKey`, it has been improved.

```go
package jwk_test

import (
	"context"
	"testing"

	"github.com/lestrrat-go/jwx/jwk"
)

const u = "https://www.googleapis.com/oauth2/v3/certs"

func BenchmarkVerifySet(b *testing.B) {
	ctx := context.Background()
	ar := jwk.NewAutoRefresh(ctx)
	ar.Configure(u)
	set, err := ar.Fetch(ctx, u)
	if err != nil {
		b.Fatal(err)
	}
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		_, ok := set.LookupKeyID("REPLACE_THE_REAL_KEY_ID")
		if !ok {
			b.Fatal("no key")
		}
	}
}
```

before
```
goos: linux
goarch: amd64
pkg: github.com/lestrrat-go/jwx/jwk
cpu: 11th Gen Intel(R) Core(TM) i3-1115G4 @ 3.00GHz
BenchmarkVerifySet-4   	 1214281	       982.7 ns/op	     304 B/op	       7 allocs/op
PASS
ok  	github.com/lestrrat-go/jwx/jwk	2.028s
```

after
```
goos: linux
goarch: amd64
pkg: github.com/lestrrat-go/jwx/jwk
cpu: 11th Gen Intel(R) Core(TM) i3-1115G4 @ 3.00GHz
BenchmarkVerifySet-4   	14858936	        71.80 ns/op	       0 B/op	       0 allocs/op
PASS
ok  	github.com/lestrrat-go/jwx/jwk	1.296s
```

### DIFF
--- a/jwe/headers.go
+++ b/jwe/headers.go
@@ -70,8 +70,7 @@ func (h *stdHeaders) Clone(ctx context.Context) (Headers, error) {
 }
 
 func (h *stdHeaders) Copy(ctx context.Context, dst Headers) error {
-	for iter := h.Iterate(ctx); iter.Next(ctx); {
-		pair := iter.Pair()
+	for _, pair := range h.makePairs() {
 		if err := dst.Set(pair.Key.(string), pair.Value); err != nil {
 			return errors.Wrapf(err, `failed to set header`)
 		}
--- a/jwe/headers_gen.go
+++ b/jwe/headers_gen.go
@@ -675,12 +675,9 @@ LOOP:
 }
 
 func (h stdHeaders) MarshalJSON() ([]byte, error) {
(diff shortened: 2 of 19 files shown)


johejo:
Oops, my benchmark func name is not good...  should be BenchmarkLookupKey....

lestrrat:
Ah yes, I incorporated that in some areas, but not all. Makes sense.

PR #470, opened 2021-09-17
[jwt] Add option for users to make JWT verification infer the `alg` field from JWKS

lestrrat:
fixes #469 
closes #465

PR #472, opened 2021-09-21
Infer JWK "alg" from cert chain

### LINKED ISSUE #469: [jwt] Add ability to infer `alg` value using heuristics
It has come to my attention that multiple people are stumbling on Microsoft Azure JWKs not providing the `alg` field.
And when a JWKS is used to perform JWT payload verification, alg is now a requirement, things go haywire in such situations.

This may be happening more than I imagined, so I'm inclined to add an option to automatically infer the algorithm used
by examining the given JWK in a JWKS.

jpJuni0r:
### Abstract

The "alg" fields for JWKs is optional. This change tries to infer its value from the certificate chain.

### Notes

I just saw that another PR (#470 by @lestrrat) resolves the same issue.

Motivation for this patch is exactly the reason from the original issue — to parse JWKs from Microsoft.

Fixes #469.

### DIFF
--- a/internal/jwxtest/jwxtest.go
+++ b/internal/jwxtest/jwxtest.go
@@ -8,12 +8,16 @@ import (
 	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/rsa"
+	"crypto/x509"
+	"crypto/x509/pkix"
 	"encoding/json"
 	"io"
 	"io/ioutil"
+	"math/big"
 	"os"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/lestrrat-go/jwx/internal/ecutil"
 	"github.com/lestrrat-go/jwx/jwa"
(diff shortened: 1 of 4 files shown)


jpJuni0r:
Your implementation seems more mature, but I'm afraid there's still a bug in it. When validating my JWT from Microsoft the parseBytes function in jwt.go finds the right JWK, but further down below you iterate over all JWKs and don't actually use the key found previously.

https://github.com/lestrrat-go/jwx/blob/da3f6f1321e67e3460161b4b2f8d8a67ddf6140e/jwt/jwt.go#L216-L219

This results in an `"kid" fields do not match` error message when calling the parse function on line 235.

lestrrat:
You mean I don't stop after I find a valid key? I haven't looked at the code yet, but that's all too possible :)
If that's the case that would seems pretty clear, but if you want to make sure I fix it (without doing a back-and-forth), can you produce a test case?

jpJuni0r:
Upon further investigation it seems like I mixed something up on my side. Your implementation now works well with my use-case :tada:.

I guess I can close this PR now, but thank you again for your help!

PR #476, opened 2021-09-23
Use index based loop for InferAlgorithmFromKey

johejo:
related #468 #470


lestrrat:
I have a pending PR #475 which changes this bit quite significantly, so I don't think this will be necessary

PR #477, opened 2021-09-24
Add option to pass arbitrary validation code

lestrrat:
Meanwhile also convert most everything in `jwt.Validate()` into `jwt.Validator` based implementation

### DIFF
--- a/jwt/validate.go
+++ b/jwt/validate.go
@@ -97,6 +97,7 @@ type isInTimeRange struct {
 	less bool // if true, d =< c1 - c2. otherwise d >= c1 - c2
 }
 
+// MaxDeltaIs implements the logic behind `WithMaxDelta()` option
 func MaxDeltaIs(c1, c2 string, dur time.Duration) Validator {
 	return &isInTimeRange{
 		c1:   c1,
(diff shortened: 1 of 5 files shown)


jwt/validate.go
```diff
@@ -44,145 +44,282 @@ func timeClaim(t Token, clock Clock, c string) time.Time {
...
-			claim := o.Value().(claimValue)
-			claimValues[claim.name] = claim.value
+			validators = append(validators, v)
 		}
 	}
 
-	for c := range requiredMap {
-		if _, ok := t.Get(c); !ok {
-			return errors.Errorf(`required claim %s was not found`, c)
+	ctx := SetValidationCtxSkew(context.Background(), skew)
```
johejo, line 82: Is it possible to propagate the parent context?

lestrrat: hmmm, no, because `jwt.Validate()` does not take a context :/

johejo: Is there any new API?

```go
jwt.Validate(
  jwt.ValidateWithContext(ctx),
)
```

Or make jwt.Validate depreacated
```go
jwt.ValidateWithContext(ctx, ...)
```

lestrrat: Sure, it's _doable_ but why? is there a usecase?

johejo: I considered some usecases for claim validation logic.
- RPC call to another server over network
- Add span with tracing framework such as OpenTelemetry

lestrrat: Ah, the span, I can sympathize with...

lestrrat: fixed in bb7f584

johejo (commented):
LGTM

Issue #500, opened 2021-11-21
JWT Constructor could take variable slice of Options to build token with


ddouglas:
**Abstract**
This is possible a refactor of `jwt.New()`, I haven't really dug into what is does yet, but the idea is that like `jwt.Parse` and `jwt.Validate`, `jwt.New` could take a slice of options that allow the user to set various claims on the jwt before the jwt is returned.

**Describe the proposed solution/change**
I'm not sure right now, this is a brain dump and since you have a feature list for V2, I'd like to see if this could be added, however I do have some pesudo code

Right now, my token initialization for a user after they login looks like this.
```
	token := jwt.New()

	err = token.Set(jwt.AudienceKey, "pillbox")
	if err != nil {
		return "", fmt.Errorf("failed to set audience on jwt: %w", err)
	}

	err = token.Set(jwt.IssuerKey, "pillbox")
	if err != nil {
		return "", fmt.Errorf("failed to set issuer on jwt: %w", err)
	}

	err = token.Set(jwt.IssuedAtKey, time.Now().Unix())
	if err != nil {
		return "", fmt.Errorf("failed to set issued at on jwt: %w", err)
	}

	err = token.Set(jwt.ExpirationKey, time.Now().Add(time.Hour*6).Unix())
	if err != nil {
		return "", fmt.Errorf("failed to set expiration on jwt: %w", err)
	}

	err = token.Set("owner_id", owner.ID.String())
	if err != nil {
		return "", fmt.Errorf("failed to set owner_id on jwt: %w", err)
	}

	err = token.Set("email", owner.Email)
	if err != nil {
		return "", fmt.Errorf("failed to set email on jwt: %w", err)
	}

	keyBytes, err := jwt.Sign(token, jwa.RS256, s.privateJwk)
	if err != nil {
		return "", fmt.Errorf("failed to generate jwk set: %w", err)
	}
```

This could be simplify in two ways. One, by my lazy a** actually making the changes I said I would make in #498 and submit a PR, followed by the request in this issue

```
token, err := jwt.New(
	jwt.WithClaim(jwt.ClaimField, "claimValue"), 
	jwt.WithClaim(jwt.ClaimField, "claimValue"),
	jwt.WithClaim(jwt.ClaimField, "claimValue"),
		jwt.WithClaim(jwt.ClaimField, "claimValue"),
)

signed, err := jwt.Sign(....)
if err != nil {
	return err
}

return string(signed), err
```

[...]

PR #501, opened 2021-11-22
[jwt] Introduce Builder pattern for JWT package

lestrrat:
refs #500


lestrrat:
ATM I do not like the name NewWith, nor the WithClaim option, but it's a start.

Also, currently there's a problem with providing an option-based interface to jwt/openid package. I don't know what the best approach is yet.

Random ideas

Builder. Easier programmatic generation. Slightly ugly

```go
tok, err := jwt.Builder().
   Claim(foo, fooValue).
   Claim(bar, barValue).
   Build()
```

public option identifier functions so jwt/openid can figure out private options:

```go
package jwt

func IsFooOption(o Option) bool { ... }

package openid

   switch {
   case jwt.IsFooOption(option):
      ....
    }
}
```

Other random API ideas

```go
jwt.Contruct(jwt.Claim(foo, fooValue), jwt.Claim(bar, barValue))
```

lestrrat:
Hmmm, ATM I like the Builder pattern.

```go
tok, err := jwt.NewBuilder().
  Claim(...).
  Claim(...).
  Build()
```

PR #507, opened 2021-12-06
Add deprecation message to lestrrat/go-jwx go.mod

zrhoffman:
(PRing here because [github.com/lestrrat/go-jwx](github.com/lestrrat/go-jwx) is archived and closed to PRs)

Although the `github.com/lestrrat/go-jwx` README does mention that this repo has been moved, if a user goes to [pkg.go.dev/github.com/lestrrat/go-jwx](https://pkg.go.dev/github.com/lestrrat/go-jwx), pkg.go.dev does not know to display a deprecation message in the top-left corner ([example](https://pkg.go.dev/github.com/golang/protobuf)), because there is no Go modules [deprecation message](https://go.dev/ref/mod#go-mod-file-module-deprecation). This PR adds `go.mod`/`go.sum` and includes a deprecation message for module users.

Users will also be told that `github.com/lestrrat/go-jwx` is deprecated, and to use `github.com/lestrrat-go/jwx` instead, if they try to `go get` it.


lestrrat:
Thanks. I think I did the necessary bit in the original repo, but am waiting for changes to propagate in pkg.go.dev

lestrrat:
Done. Closing this PR as I have already did what's necessary in the original repo
image

PR #513, opened 2021-12-14
[jws] Add ability to verify payload using "jku"

### LINKED ISSUE #512: [jws] Automatically respect JWKS in `jku` to verify the JWS payload
Probably something like `jws.Verify(..., jws.WithAutoFetchJKU(true))`.

Because this would be fetching the remote resource, we must be aware of security impact.
We probably should document it is highly recommended to ONLY use this option with a whitelist
(also, we would have to create wrapper options? to pass `jwk.FetchOption` to `jws.Verify()`)

lestrrat:
fixes #512

### DIFF
--- a/Changes
+++ b/Changes
@@ -11,7 +11,7 @@ v1.2.14 UNRELEASED
     are different, and 2) Users MUST be aware that they are doing a totally
     different operation than a regular `jws.Verify()`
 
-[Buf fixes]
+[Bug fixes]
   * `jws.SignMulti()` has been fixed to assign the "kid" field of the key used
     for signing the payload
   * `jws.SignMulti()` has been fixed to respect the "kid" field of the protected
--- a/Changes
+++ b/Changes
@@ -5,6 +5,17 @@ v1.2.14 UNRELEASED
 [New Features]
   * `jwk.Fetch()` and `(*jwk.AutoRefresh).Configure()` can now take `jwk.Whitelist`
     object to check for the validity of a url to be fetched
+  * `jws.VerifyAuto()` has been added to verify payloads that can be verified
+    using the JWK set provided in the "jku" field. This function is purposely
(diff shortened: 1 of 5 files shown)


Changes
```diff
@@ -5,6 +5,17 @@ v1.2.14 UNRELEASED
 [New Features]
   * `jwk.Fetch()` and `(*jwk.AutoRefresh).Configure()` can now take `jwk.Whitelist`
     object to check for the validity of a url to be fetched
+  * `jws.VerifyAuto()` has been added to verify payloads that can be verified
+    using the JWK set provided in the "jku" field. This function is purposely
+    separated from the `jws.Verify()` function because 1) the required parameters
+    are different, and 2) Users MUST be aware that they are doing a totally
+    different operation than a regular `jws.Verify()`
+
+[Buf fixes]
```
gonzaloserrano, line 14:
```suggestion
[Bug fixes]
```

lestrrat: oops, thanks! :D

jws/jws.go
```diff
@@ -312,12 +365,37 @@ func verifyJSON(signed []byte, alg jwa.SignatureAlgorithm, key interface{}, dst
...
+			}
+
+			verifier, err := NewVerifier(ctx.alg)
+			if err != nil {
+				return nil, errors.Wrap(err, "failed to create verifier")
+			}
+
+			if _, err := ctx.tryVerify(verifier, sig.protected, buf.Bytes(), sig.signature, m.payload); err == nil {
+				if ctx.dst != nil {
+					*(ctx.dst) = m
```
gonzaloserrano, line 384: not that I have any knowledge about how this works but I noticed this line and L393 are not covered by tests

lestrrat: Thanks, addressed in 7a2ff0c

PR #515, opened 2021-12-16
[jwt] Add options to enable verification via jws.VerifyAuto()

### LINKED ISSUE #514: [jwt] Make it possible to call jws.VerifyAuto from jwt.Parse

lestrrat:
fixes #514


lestrrat:
As of 1ccba3a the default behavior of `jws.VerifyAuto()` is to not allow any URLs to be fetched. This is to force `jwt.Parse()` users who wish to use `jws.VerifyAuto()` underneath think about whitelisting URLs that can be fetched via jku.

PR #523, opened 2021-12-28
jwt: only infer algorithm on the already-selected key

imirkin:
There's no issue in practice, I think, but it's silly to loop over the keys when we've already selected the key up front...

`jws.verifyCompact` looks like it makes sure that the kid's match, but it's a bit unnecessary to rely on that here.


lestrrat (approved):
LGTM. Thanks!

PR #570, opened 2022-02-15
Allow processing incoming payload before parse

lestrrat:


lestrrat:
closed in favor of #574

PR #571, opened 2022-02-15
Fix typos

rodrigc:


lestrrat:
Wow, these went unnoticed for a long time... Thanks!

PR #574, opened 2022-02-16
Add IgnoreParseError option

lestrrat:


jwk/set.go
```diff
@@ -225,7 +227,10 @@ LOOP:
 				for i, keysrc := range list {
 					key, err := ParseKey(keysrc, options...)
 					if err != nil {
-						return errors.Wrapf(err, `failed to code key #%d in "keys"`, i)
+						if !ignoreParseError {
+							return errors.Wrapf(err, `failed to decode key #%d in "keys"`, i)
```
mzc-lunar, line 231: One thing I thought when seeing this is that I guess it could be considered doing a log that the parsing failed, just in case the consumer would want that information.

But I don't think the library really logs anything in general, right?

lestrrat: Nope, no logging goes in jwx. This is a trade-off: you get to ignore errors and lose the errors, but you can't ignore _AND_ get the errors. I expect this to be only used in extreme cases, and if you need to keep watching for the validity of the JWKS, I'd say you should have a separate monitoring process that doesn't ignore errors.

PR #580, opened 2022-02-23
Change text in error message to base64url encoding

rodrigc:
I would suggesting changing the text in this error message to `base64url encoding`.

to match the terminology in RFC7515
https://datatracker.ietf.org/doc/html/rfc7515#section-2

`base64` and `base64url` are not exactly the same.  For people trying to diagnose problems by reading this error message, and not 100% familiar with JWX, they may go down the wrong path of trying to use `base64` decoders for solving their problem instead of a `base64url` decoder.  (I did).

The code internally uses `internal/base64/base64.go` for encoding/decoding which already implements base64url.

### DIFF
--- a/internal/json/json.go
+++ b/internal/json/json.go
@@ -34,7 +34,7 @@ func AssignNextBytesToken(dst *[]byte, dec *Decoder) error {
 
 	buf, err := base64.DecodeString(val)
 	if err != nil {
-		return errors.Errorf(`expected base64 encoded []byte (%T)`, val)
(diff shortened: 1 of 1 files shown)


lestrrat:
Technically, you're not correct either. In the wild,  implementations that (gasp) use various different base64 types exist, and that's why the internal/base64 package exists. Otherwise I get people asking me why they can't decode their objects.The base64.DecodeString uses simple heuristics to decide which encoding we should use -- so the if you want to be pedantic about the error message it should read `"base64 or base64url"`, I think

PR #613, opened 2022-03-05
Handroll autodocs

lestrrat:
We need to keep documentation working


lestrrat:
going to merge this batch for now

PR #624, opened 2022-03-07
docs: update jwt documentation with section explain use of jwk.fetch …

akoserwal:
…to parse & validate jwt token


lestrrat:
Thanks, but I'm a bit reluctant to put this as is.

First, there already exists documentation on [JWK fetching else where](https://github.com/lestrrat-go/jwx/blob/main/docs/04-jwk.md#fetching-jwk-sets). I think it would be better to include a link to that section in the appropriate section for verifying with JWKS.

Second,  I don't think you can claim that JWKS will always reside in those URLs. At least you would have to qualify that they are _usually_ found there, but not necessarily so.

akoserwal:
I totally agree with you.
1) Linking section [JWK fetching else where](https://github.com/lestrrat-go/jwx/blob/main/docs/04-jwk.md#fetching-jwk-sets). here as well
2) the First section explains it well, it can be removed.

lestrrat:
I'm going to close this for now because I'm currently working on [v2](https://github.com/lestrrat-go/jwx/tree/v2) stuff (which already addresses this issue), and would like to focus my efforts there. If you would like to redo the PR and address the above issues, feel free to reopen. Thanks!

PR #654, opened 2022-03-23
Deprecate jwk.AutoRefresh and replace it with a shim over httprc.Cache for v2

### LINKED ISSUE #652: Remove jwk.AutoRefresh and make it a standalone module
in anticipation of v1 and v2 diverging, I think we should decouple it and create a `jwkcache` module

lestrrat:
By removing the bulk of the logic out to a separate library, we can
bring improvements and fixes to multiple versions at the same time

fixes #652


imirkin:
Fine by me!

shaunco:
Looks good to me! Really appreciate all the effort you put in to JWX!

PR #686, opened 2022-04-13
Fix race in `jwk.AutoRefresh`

sergicastro:
Using the library I think I found a race when `AutoRefresh.Configure` and `AutoRefresh.doRefreshRequests` happen at the same time because both are accessing the `jwk.target` struct.

I added the first commit with a test that shows up the race, and then a commit that fixes this race.
I've been using the same locker I saw it is used in other points of the code, but maybe the best approach would be to have a locker for each instance of `jwk.target`.

### DIFF
--- a/jwk/refresh.go
+++ b/jwk/refresh.go
@@ -489,9 +489,9 @@ func (af *AutoRefresh) refreshLoop(ctx context.Context) {
 func (af *AutoRefresh) doRefreshRequest(ctx context.Context, url string, enableBackoff bool) error {
 	af.muRegistry.RLock()
 	t, ok := af.registry[url]
-	af.muRegistry.RUnlock()
 
 	if !ok {
+		af.muRegistry.RUnlock()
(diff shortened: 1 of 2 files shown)


sergicastro:
@lestrrat  I installed go 1.16.x in my machine I run several times the command `make cover-goccy`, and even I used `-count` flag to identify if the test is flaky.

But I'm not able to make it fail, could you give me a hand so I can understand what's happening with that failed check?

lestrrat:
@sergicastro Thank you! Let me think about this one. It might just be better if I could replace the current code with github.com/lestrrat-go/httprc

lestrrat:
@sergicastro That's a flaky test that I've been meaning to fix. I know it's not great, but don't worry about it for now

lestrrat:
For the record, I did a brief debugging session, and also pondered about httprc, but I think it would take a bit more time to come up with a solution, so we'll take the proposed fix until next time we have more time to debug/redesign throroughly

PR #718, opened 2022-04-27
Implement custom validation errors for invalid `aud` and `iss` claims

ste93cry:
As discussed in #706, I'm adding two new custom validation errors for when the `aud` and `iss` claims are invalid. Do you think that the `Validate()` functions should always return the specific error, even in case the claim does not exists?

### DIFF
--- a/docs/01-jwt.md
+++ b/docs/01-jwt.md
@@ -743,7 +743,7 @@ func ExampleJWT_ValidateIssuer() {
   }
   fmt.Printf("%s\n", err)
   // OUTPUT:
-  // "iss" not satisfied
+  // "iss" not satisfied: values do not match
 }
 ```
 source: [examples/jwt_validate_issuer_example_test.go](https://github.com/lestrrat-go/jwx/blob/v2/examples/jwt_validate_issuer_example_test.go)
--- a/docs/01-jwt.md
+++ b/docs/01-jwt.md
@@ -743,7 +743,7 @@ func ExampleJWT_ValidateIssuer() {
   }
   fmt.Printf("%s\n", err)
   // OUTPUT:
-  // "iss" not satisfied: values do not match
+  // "iss" not satisfied
 }
 ```
 source: [examples/jwt_validate_issuer_example_test.go](https://github.com/lestrrat-go/jwx/blob/v2/examples/jwt_validate_issuer_example_test.go)
--- a/jwt/options.go
+++ b/jwt/options.go
@@ -167,7 +167,7 @@ func WithKeySet(set jwk.Set, options ...interface{}) ParseOption {
 // WithIssuer specifies that expected issuer value. If not specified,
 // the value of issuer is not verified at all.
 func WithIssuer(s string) ValidateOption {
-	return WithValidator(IssuerClaimValueIs(s))
+	return WithValidator(issuerClaimValueIs(s))
 }
 
 // WithSubject specifies that expected subject value. If not specified,
@@ -187,7 +187,7 @@ func WithJwtID(s string) ValidateOption {
 // matches this value.  If not specified, the value of issuer is not
 // verified at all.
 func WithAudience(s string) ValidateOption {
-	return WithValidator(AudienceClaimContainsString(s))
+	return WithValidator(audienceClaimContainsString(s))
 }
 
 // WithClaimValue specifies the expected value for a given claim
--- a/jwt/options.go
+++ b/jwt/options.go
@@ -167,7 +167,7 @@ func WithKeySet(set jwk.Set, options ...interface{}) ParseOption {
 // WithIssuer specifies that expected issuer value. If not specified,
 // the value of issuer is not verified at all.
 func WithIssuer(s string) ValidateOption {
-	return WithValidator(ClaimValueIs(IssuerKey, s))
+	return WithValidator(IssuerClaimValueIs(s))
 }
 
 // WithSubject specifies that expected subject value. If not specified,
@@ -187,7 +187,7 @@ func WithJwtID(s string) ValidateOption {
 // matches this value.  If not specified, the value of issuer is not
 // verified at all.
 func WithAudience(s string) ValidateOption {
-	return WithValidator(ClaimContainsString(AudienceKey, s))
+	return WithValidator(AudienceClaimContainsString(s))
 }
 
 // WithClaimValue specifies the expected value for a given claim
--- a/jwt/validate.go
+++ b/jwt/validate.go
@@ -178,6 +178,8 @@ func (*missingRequiredClaimError) Unwrap() error          { return nil }
 var errTokenExpired = NewValidationError(fmt.Errorf(`"exp" not satisfied`))
 var errInvalidIssuedAt = NewValidationError(fmt.Errorf(`"iat" not satisfied`))
 var errTokenNotYetValid = NewValidationError(fmt.Errorf(`"nbf" not satisfied`))
+var errTokenInvalidAudience = NewValidationError(fmt.Errorf(`"aud" not satisfied`))
+var errTokenInvalidIssuer = NewValidationError(fmt.Errorf(`"iss" not satisfied`))
 
 // ErrTokenExpired returns the immutable error used when `exp` claim
 // is not satisfied
@@ -193,6 +195,14 @@ func ErrInvalidIssuedAt() ValidationError {
 
 func ErrTokenNotYetValid() ValidationError {
 	return errTokenNotYetValid
+}
+
+func ErrTokenInvalidAudience() ValidationError {
+	return errTokenInvalidAudience
+}
+
+func ErrTokenInvalidIssuer() ValidationError {
+	return errTokenInvalidIssuer
 }
 
 // ErrMissingRequiredClaim creates a new error for missing required claims.
@@ -329,7 +339,7 @@ func ClaimContainsString(name, value string) Validator {
 // IsValidationError returns true if the error is a validation error
 func IsValidationError(err error) bool {
 	switch err {
-	case errTokenExpired, errTokenNotYetValid, errInvalidIssuedAt:
+	case errTokenExpired, errTokenNotYetValid, errInvalidIssuedAt, errTokenInvalidAudience, errTokenInvalidIssuer:
 		return true
 	default:
 		switch err.(type) {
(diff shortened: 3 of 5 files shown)


jwt/options.go
```diff
@@ -167,7 +167,7 @@ func WithKeySet(set jwk.Set, options ...interface{}) ParseOption {
 // WithIssuer specifies that expected issuer value. If not specified,
 // the value of issuer is not verified at all.
 func WithIssuer(s string) ValidateOption {
-	return WithValidator(ClaimValueIs(IssuerKey, s))
+	return WithValidator(IssuerClaimValueIs(s))
```
lestrrat, line 170: I don't think it's correct to add an exported function here.
The public API to check `iss` is already there (`WithIssuer`). Its internals need not be public.

jwt/options.go
```diff
@@ -187,7 +187,7 @@ func WithJwtID(s string) ValidateOption {
 // matches this value.  If not specified, the value of issuer is not
 // verified at all.
 func WithAudience(s string) ValidateOption {
-	return WithValidator(ClaimContainsString(AudienceKey, s))
+	return WithValidator(AudienceClaimContainsString(s))
```
lestrrat, line 190: ditto

jwt/validate.go
```diff
@@ -195,6 +197,14 @@ func ErrTokenNotYetValid() ValidationError {
 	return errTokenNotYetValid
 }
 
+func ErrTokenInvalidAudience() ValidationError {
```
lestrrat, line 200: In hindsight, `errTokenNotYetValid` was a misnomer which should have been corrected in 2.0.0...
This should also be `ErrInvalidAudience()`, as well as the error value itself.

jwt/validate.go
```diff
@@ -195,6 +197,14 @@ func ErrTokenNotYetValid() ValidationError {
...
+	return errTokenInvalidAudience
+}
+
+func ErrTokenInvalidIssuer() ValidationError {
```
lestrrat, line 204: ditto

jwt/validate.go
```diff
@@ -352,17 +362,41 @@ func (ccs claimContainsString) Validate(_ context.Context, t Token) ValidationEr
...
+// AudienceClaimContainsString can be used to check if the audience claim, which is
+// expected to be a list of strings, contains `value`.
+func AudienceClaimContainsString(value string) Validator {
+	return audienceClaimContainsString{value: value}
+}
+
+func (ccs audienceClaimContainsString) Validate(_ context.Context, t Token) ValidationError {
+	v, ok := t.Get(AudienceKey)
+	if !ok {
+		return NewValidationError(fmt.Errorf(`claim %q not found`, AudienceKey))
```
lestrrat, line 386: I don't understand why this is not of type `errInvalidAudience`. If the claim doesn't exist, then the audience value is invalid, right?

jwt/validate.go
```diff
@@ -389,6 +423,27 @@ func (cv *claimValueIs) Validate(_ context.Context, t Token) ValidationError {
...
+}
+
+type issuerClaimValueIs struct {
+	value string
+}
+
+func (cv *issuerClaimValueIs) Validate(_ context.Context, t Token) ValidationError {
+	v, ok := t.Get(IssuerKey)
+	if !ok {
+		return NewValidationError(fmt.Errorf(`%[1]q not satisfied: claim %[1]q does not exist`, IssuerKey))
```
lestrrat, line 439: Same as audience error. I think this should be a `errIssuerInvalid`

docs/01-jwt.md
```diff
@@ -743,7 +743,7 @@ func ExampleJWT_ValidateIssuer() {
   }
   fmt.Printf("%s\n", err)
   // OUTPUT:
-  // "iss" not satisfied: values do not match
```
lestrrat, line 746: This doesn't hurt, but ... why? I don't know, it doesn't strike me as something that's important

lestrrat (changes requested):

lestrrat:
>  Do you think that the Validate() functions should always return the specific error, even in case the claim does not exists?

There's a nuance to this statement that (I don't know if you intended to be that way) that I don't agree with.
Specifically, "Validate() ... should _always_ return the specific error" doesn't sound right. But for the purposes of this PR, yes, I think it should return the specific error.

It's not really that `Validate()` needs to _always_ return a particular value, but it has more to do with the user's expectation. If I were a user and asked my `iss` claim to be validated, and wrote `jwt.Validate(tok, jwt.WithIssuer("foobar")`, I would expect `iss == "foobar"` to be checked all the time. Therefore, empty values should error out.

There may be cases where empty/non-existent values can be ignored, but not in this case.

docs/01-jwt.md
ste93cry: I changed the message just to be consistent with the other similar errors. However, since the `Validate()` function is expected to return always the same error type with a different message depending on the case (existence of the claim, invalid value, etc), I can definitely revert the change

jwt/validate.go
ste93cry: Yes, this actually answer the question I asked in the PR's description. I agree that it should return `errInvalidAudience`, but wanted to be sure before making more changes

ste93cry:
>But for the purposes of this PR, yes, I think it should return the specific error

What I meant is that I had the doubt if a generic `ValidationError` had to be returned for empty/non-existent values or if the specific `ErrXXX` had to be (for these particular cases of course), but you answered to me in https://github.com/lestrrat-go/jwx/pull/718#discussion_r860274458 and I agree with your thoughts

jwt/validate.go
```diff
@@ -175,6 +175,34 @@ func (err *missingRequiredClaimError) Is(target error) bool {
 func (err *missingRequiredClaimError) isValidationError() {}
 func (*missingRequiredClaimError) Unwrap() error          { return nil }
 
+type invalidAudienceError struct {
```
lestrrat, line 178: (Note, this is for me, you don't have to do anything @ste93cry )
This should just be generated so it's easier to add errors in the future

jwt/validate_test.go
```diff
@@ -134,7 +158,7 @@ func TestGHIssue10(t *testing.T) {
 			if !assert.Error(t, err, "token.Validate should fail") {
 				return
 			}
-			if !assert.False(t, errors.Is(err, jwt.ErrTokenNotYetValid()), `error should be not ErrNotYetValid`) {
+			if !assert.ErrorIs(t, err, jwt.ErrInvalidAudience(fmt.Errorf("")), `error should be ErrInvalidAudience`) {
```
lestrrat, line 161: Untested, but I _think_ you can just compare it against `jwt.ErrInvalidAudience(nil)`

And if that works, I think you can do this:

```go
func ErrInvalidAudience() ValidationError { // does not take argument
  return &errInvalidAudience{} // no error set
}
```

ste93cry: So, are you proposing that I always instantiate the struct where needed so that I can specify the message, while people from the outside have to use the function and aren't required to pass the error?

lestrrat: Sorry, that was a fat finger. Let me try this again in my next post

lestrrat: What I want to do is to NOT expose the `error` parameter to the users: i.e. want `jwt.ErrInvalidAudience()`, not `jwt.ErrInvalidAudience(fmt.Errorf(...))` or `jwt.ErrInvalidAudience(nil)`

To allow that, this test needs to pass (I'm only showing the struct instantiation so you can see the comparison that I want to make):

```go
// definition
type invalidAudienceError struct {
  err error
}

// I think errors.Is returns true even if you use an empty struct
if !assert.ErrorIs(t, err, jwt.invalidAudience{err: nil}) { }
```

Note that I haven't ran tests with it so I'm not sure if the above holds. if the above doesn't work, the rest of my post is useless. But I think it works.
If the above works, then you should be able to change your code to do this:

```go
type invalidAudienceErr struct {
  err error
}
var errInvalidAudience = &invalidAudienceErr{} // empty err

func ErrInvalidAudience() error {
   return errInvalidAudience
}

// in your test
if !assert.ErrorIs(t, err, jwt.ErrInvalidAudience(), ....) { }
```

As a result:
* Users see ErrInvalidAudience with a simpler interface (if you were a user, wouldn't you wonder why you need to pass an extra argument just to compare the returned error?)
* ErrInvalidAudience() will always return the same instance of the error, and it's more efficient.

lestrrat: Also, you may not have to provide `Unwrap()` to `invalidAudienceErr`... I think. Again, unconfirmed.

ste93cry: Ok, I understood well. I agree on changing this, the obvious downside is that I need to instantiate the `struct` everywhere, therefore bypassing what I would define a "constructor". Anyway, the tradeoff is worth a simpler and clearer API 🎉

lestrrat: Not sure about `instantiate struct everywhere`. I'm obviously not the one writing this code, but at first glance I don't think you would be instantiating any _extra_ errors.

ste93cry: I meant that where I was using `ErrInvalidAudience(...)` before in the internal code I now have to write `&invalidAudienceError{...}`. Yea, it's just one place so it's fine, but since this is basically bypassing the official constructor of the `struct` it may be risky in the future if things change. But, as I said, I agree that this makes the API a lot more clearer

lestrrat: Oh, I see. You were writing `return ErrInvalidAudience(subErr)`, and you are saying now you have to write `return invalidAudienceError{err: subErr}`.

If that really bothers you, I'm fine wit you impementing `newInvalidAudienceError(error)`, but I don't think it pays off in this particular case because there is only 1 location where this is done. 


---

This is an aside, so I'm definitely not asking you to make this change, but since in Go you can't enforce use of a constructor within the same package, you will need to separate the errors into a another package to really make the constructors enforceable, e.g.

```go
// internal/errors/errors.go (or similar)
type invalidAudienceError { ... }

func NewInvalidAudience(err error) {
   return &invalidAudienceError{err: err}
}

// jwt/validate.go
import "github.com/lestrrat-go/jwx/v2/internal/errors"

var errInvalidAudience = errors.NewInvalidAudience(nil)
func ErrInvalidAudience() ValidationError {
  return errInvalidAudience{}
}

func validateAudience() ... {
   if audienceIsInvalid {
     return errors.NewInvalidAudience(subErr)
   }
}
```

If you don't do this, developers are free to just write `&invalidAudienceErr{}` anywhere they please.

lestrrat (approved):
LGTM

jwt/validate_test.go
ste93cry: >I don't think it pays off in this particular case because there is only 1 location where this is done

Yup, agree. Change done, but I just found out that this code will `panic`:

```go
err := jwt.ErrInvalidAudience()
_ = err.Error()
```

I think that implementing `Error()` to return an empty `string` will solve the issue. What do you think?

lestrrat: Ah... While I think for efficiency's sake an empty string works, it becomes a problem when users try to stringify the empty error:

```go
fmt.Printf("%s", jwt.ErrInvalidAudience)
```

so... maybe

```go
func (err invalidAudienceErr) Error() string {
  if err.err == nil {
     return "jwt: invalid audience"
  }
  return err.err.Error()
}
```

or something similar to

```go
func (err invalidAudienceErr) Error() string {
  return fmt.Errorf(`jwt: invalid audience: %s`, err.err)
}
```

ste93cry: >it becomes a problem when users try to stringify the empty error

Good point! Maybe we could just instantiate the `struct` with a default error, instead of implementing the `Error()` method ourselves?

lestrrat: ATM I think I prefer implementing an `Error()`. I know there are pros/cons there, because
1. it just irks me that there is a glaring panic
2. I don't want to go the route of a full-fledged error for something that is only used a couple of times

ste93cry: Done. Last doubt: 

> >Also, you may not have to provide `Unwrap()` to `invalidAudienceErr`
>
> It works, I tried that already. The reason I chose to call that method was that I didn't want an error that is a validation error to wrap another validation error

Since I edited the message while you were posting yours, I'm not sure you noticed the change. Are we good to go, or should I change this?

lestrrat: Nah, it doesn't hurt. You can leave it :)

lestrrat (approved):
LGTM

PR #720, opened 2022-04-28
Fix `jwt.isValidationError()` false negative for `ErrMissingRequiredClaim`

### LINKED ISSUE #719: `jwt.isValidationError()` false negative for `ErrMissingRequiredClaim`
**Describe the bug**

When implementing the `missingRequiredClaimError` `struct`, the `isValidationError()` function was not updated to take into account the new error. I though it was working because of the `err.(*validationError)` cast, but I forgot that the `struct` is not an instance of that.

`go version go1.18.1 windows/amd64`

```go
if !jwt.IsValidationError(jwt.ErrMissingRequiredClaim("...")) {
    panic("not a validation error") // not expected outcome
}
```

ste93cry:
Fixes #719: I simply added a new case to detect whether the target error is an instance of `missingRequiredClaimError` and changed the test to avoid a regression in the future

### DIFF
--- a/jwt/jwt_test.go
+++ b/jwt/jwt_test.go
@@ -1292,8 +1292,12 @@ func TestGH430(t *testing.T) {
 }
 
 func TestGH706(t *testing.T) {
-	tok := jwt.New()
-	if !assert.ErrorIs(t, jwt.ErrMissingRequiredClaim(""), jwt.Validate(tok, jwt.WithRequiredClaim("foo")), `jwt.Validate should fail`) {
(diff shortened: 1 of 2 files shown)


lestrrat (approved):

lestrrat:
Thanks!

ste93cry:
Just to understand when I can close the loop downstream, do you have an ETA for the next release?

lestrrat:
not ETA yet. I usually let these sit for a week or so

PR #742, opened 2022-05-17
fix typos

satorunooshie:
I found a couple of typos using this amazing library so I fixed it.
I wonder I don't have to fix typos under `docs/`.


lestrrat:
Thanks. I'm sure there are many typos all over. Let me know if you find any.

PR #758, opened 2022-06-15
Accept multiple keys given a key ID

lestrrat:
Builds on top of #757 
* Refactor and add tests
* Refactor code
* Add `WithMultipleKeysPerKeyID` to preserve old behavior

I think the last bit is necessary because we should keep new behavior opt-in as much as possible so that people upgrading do not get surprised.


lestrrat:
@mike-jm thanks for #757!

I'm good with the direction of proposed changes, but I gave your PRs a few tweaks, as well as adding an explicit option to toggle the new behavior. Please take a look and let me know if this works for you.

mike-jm:
@lestrrat Thank you very much. This will work for us.

PR #773, opened 2022-07-11
fix doc bug in Key interface

muir:
I rather suspect KeyType() returns the `kty` rather than the `kid`.

### DIFF
--- a/formatkind_string_gen.go
+++ b/formatkind_string_gen.go
@@ -4,6 +4,18 @@ package jwx
 
 import "strconv"
 
+func _() {
+	// An "invalid array index" compiler error signifies that the constant values have changed.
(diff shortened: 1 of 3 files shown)


lestrrat:
Thanks!

Ah, you're right but that's a generated file, so it will be overwritten unless we change the generating code.

I know what needs to be fixed is [here](https://github.com/lestrrat-go/jwx/blob/develop/v2/tools/cmd/genjwk/main.go#L672), but I'm a bit swamped ATM, so you can either send me a PR to fix the generator, or hold on until I have a bit more free time.

lestrrat (approved):

lestrrat:
Thanks!

PR #778, opened 2022-07-19
codecov: update to v3 action

### LINKED ISSUE #743: [chore] Fix codecov
After we shuffled around the branches, codecov isn't working as expected.
I'm sure this is something silly, but I personally am not _that_ into the code coverage game and don't feel like spending time on it. It would be nice if someone knowledgeable can give me a hand.

imirkin:
v1 is deprecated and no longer works. Update to v3.

Fixes #743

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,9 +49,9 @@ jobs:
         run: make cover-${{ matrix.go_tags }}
       - name: Upload code coverage to codecov
         if: matrix.go == '1.18'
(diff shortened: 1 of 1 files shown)


imirkin:
Ah no ... looks like v1 isn't as gone as the docs make it out to be. For some reason it's looking at the wrong thing for the parent, probably needs some settings in codecov but I probably am not allowed to do it or can't figure out how. Or ... something else. In any case, no clue how to fix.

lestrrat:
Me neither! :D

PR #808, opened 2022-08-28
semi-automate updateing cmd/jwx without replace

lestrrat:


lestrrat:
nevermind. was a silly idea

PR #817, opened 2022-09-11
move internal/iter to the iter package for convenient codegen

mxfactorial:
switches dependency from internal to external iter package

depends on https://github.com/lestrrat-go/iter/pull/5


mxfactorial:
will `bash ../tools/cmd/genjwk.sh` etc if you merge iter pr5

then push

lestrrat:
I guess this just moves stuff and it's no like it will be end of the world if we go ahead but I don't think you have tried to explain why you need this? Were you trying to generate something and this gave you problems?

As a general rule of thumb, if it's just "cleanup" and nothing else, I'm a bit hesitant to make changes.

PR #866, opened 2023-02-08
Remove unused variables in ReadFile

johejo:
`var readFileOptions []ReadFileOption` are never used. （something mistake?）

Generate `var parseOptions []ParseOption` only when needs.


lestrrat:
Your code looks correct, but let me try to think (remember) if I had any reason why I did this in the first place... :)

lestrrat:
I'm going to merge this b/c 1) I can't remember why I did this for the life of me, and 2) since you were wise to not touch any of the API, this should not make any difference to the user

PR #929, opened 2023-05-26
Implement jwk.SetGlobalFetcher

### LINKED ISSUE #928: Provide a way to shutdown http cache fetcher to avoid go-routine leaks.
**Abstract**
At the moment `jwx` uses `github.com/lestrrat-go/httprc` under the hood to run a fetcher. The problem is that the `globalFetcher` instance created in the init function starts a go-routine which will run perpetually with no way to shutdown it.

We use [go-leak](https://github.com/uber-go/goleak) to check for go-routines leaks in our tests:

```go
func TestMain(m *testing.M) {
	code := m.Run()

	goleak.VerifyTestMain(m,
		goleak.IgnoreTopFunction("github.com/lestrrat-go/httprc"), //jwk has it running under the hood with no way to stop
	)

	os.Exit(code)
}
```

It would be good if `jwx` had a way to shutdown the fetcher.

`jwx` could  expose a method to do a shutdown for example `jwk.Shutdown()` which will stop the fetcher's go-routines. 

**Analysis**
Alternatively, `jwx` could provide an option to disable the fetch/http cache all together:
```
jwx.UseCaching(false)
```

lestrrat:
fixes #928

### DIFF
--- a/Changes
+++ b/Changes
@@ -23,7 +23,11 @@ v2.0.10 - UNRELEASED
     in throughput on average. It might make more difference if the input/output is large.
     If you care about this performance improvement, you should probably enable
     `goccy` JSON parser as well, by specifying `jwx_goccy,jwx_asmbase64` in your build call.
-
+  * Slightly changed the way global variables underneath `jwk.Fetch` are initialized and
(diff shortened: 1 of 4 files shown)


screwyprof (commented):
LGTM

PR #947, opened 2023-06-20
fix panic from empty seed

AdamKorcz:
`ed25519.NewKeyFromSeed()` will panic if its parameter is 0 length. Avoid the panic.

To reproduce, run this test:

```go
package main

import (
        "github.com/lestrrat-go/jwx/v2/jwk"
        "testing"
)

func TestIt(t *testing.T) {
        raw := []byte("{\"crv\":\"Ed25519\",\"d\":\"\",\"x\":\"\",\"KtY\":\"OKP\"}")
        k, err := jwk.ParseKey(raw)
        if err != nil {
                return
        }
        var exported []byte
        k.Raw(&exported)
}
```

### DIFF
--- a/jwk/okp.go
+++ b/jwk/okp.go
@@ -81,6 +81,9 @@ func (k *okpPublicKey) Raw(v interface{}) error {
 }
 
 func buildOKPPrivateKey(alg jwa.EllipticCurveAlgorithm, xbuf []byte, dbuf []byte) (interface{}, error) {
+	if len(dbuf) == 0 {
+		return nil, fmt.Errorf(`cannot use empty seed`)
+	}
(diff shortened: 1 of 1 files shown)


lestrrat:
Can you please show me how this occurs (preferably in a Go test code)?
I'm guessing this happens when you create an OKP key from a faulty JWK with the field `x` missing or faulty, but I just want to make sure the scenario and include a test case for it.

AdamKorcz:
Added a reproducer.

lestrrat:
Okay, so it's as suspected, thanks.

lestrrat:
merged as #949. Thanks!

PR #948, opened 2023-06-20
remove unnecessarily err checks

shogo82148:
godoc says hash.Hash.Write never returns an error: https://pkg.go.dev/hash#Hash

> Write (via the embedded io.Writer interface) adds more data to the running hash.
> It never returns an error.

So, we don't need to check errors of Write and panics in Hmac.ComputeAuthTag.

### DIFF
--- a/jwe/internal/aescbc/aescbc.go
+++ b/jwe/internal/aescbc/aescbc.go
@@ -140,17 +140,20 @@ func (c Hmac) Overhead() int {
 }
 
 func (c Hmac) ComputeAuthTag(aad, nonce, ciphertext []byte) ([]byte, error) {
-	buf := make([]byte, len(aad)+len(nonce)+len(ciphertext)+8)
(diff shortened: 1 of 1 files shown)


lestrrat:
I understand the removal of `hash.Write()` error check. I'm okay with that part.

But I'm uncomfortable with the removal of the defer code, because the comment says they have to do with `Open` and `Seal` potentially panic'ing (TBH while I remember encountering those panics, I don't remember the details of _why_ they happened). To me, they seemed unrelated to the `hash.Write` fix.

So, my initial thinking is that they shouldn't be included in the same PR?
Also, because at least my initial code and reasoning for adding those defers were for aead.Open and aead.Seal, can you please clarify why they are not necessary inside your commit message.

shogo82148:
OK, I reverted the defer code.

lestrrat:
Thank you!

PR #950, opened 2023-06-29
add size check

AdamKorcz:
To avoid this panic: https://github.com/golang/go/blob/3adcce5ae7203e9ee7bee93a4ad913a3145e682e/src/crypto/ed25519/ed25519.go#L168


lestrrat (approved):

lestrrat:
thanks!

PR #953, opened 2023-07-03
Add test case to #952

lestrrat:
Adds a test to #952. Did verify that the issue exists before the patch and not after the patch using this test.
supersedes and closes #952


ItalyPaleAle:
Thanks for this! I wasn't sure how to best write a test for this, and this looks great

ItalyPaleAle (approved):

lestrrat:
@ItalyPaleAle was going to let you know after I checked the CI, but I guess you already caught it :)

PR #960, opened 2023-07-19
Ensure http status is good before attempting to parse

th3fallen:
Saves some extra work if the provided url is returning bad data.


lestrrat:
@th3fallen Hi, thanks for the PR. Can you please include a test case for it in the PR?

Issue #912, opened 2023-04-28
Expose Registering new JWK types


decentralgabe:
**Abstract**
I'd like to experiment with Post Quantum signature schemes using JWKs. There are a number of IETF drafts that enable this possibility:
- https://datatracker.ietf.org/doc/draft-ietf-cose-dilithium/
- https://datatracker.ietf.org/doc/draft-ietf-cose-falcon/
- https://datatracker.ietf.org/doc/draft-ietf-cose-sphincs-plus/

**Describe the proposed solution/change**
The ask would be to expose a method similar to [RegisterCurve](https://github.com/lestrrat-go/jwx/blob/develop/v2/internal/ecutil/ecutil.go#LL19C1-L19C1) for experimental key types and signature schemes.

Additionally there would need to be support for these types in [functions like this](https://github.com/lestrrat-go/jwx/blob/develop/v2/jwk/jwk.go#L44), and complementary signing code.

I would recommend enabling this as an experimental feature, similar to the build tag solution for secp256k1 support today.

**Analysis**
Use a different library for signing with these keys/signature algorithms which would make me sad 😢 

**Additional context**
If there's a way to extend the library without forking today I missed - please let me know.

lestrrat:
I briefly looked at this before, but if I understand correctly you don't need to register a _JWK_.

It's the signer/verifier and encrypter/decrypter that needs to be [registered](https://github.com/lestrrat-go/jwx/blob/develop/v2/docs/02-jws.md#using-a-custom-signingverification-algorithm) because the object that
handles the process is triggered by looking at the algorithm name. But the jwx framework the key is
just an opaque object until the signer/verifier/encrypter/decrypter sees it, at which point the
algorithm may dictate a particular type of JWK

So as far as the original question in the title goes, you don't need to register a JWK type and there's already code to register algorithms  (but yes, apparently [we have a bug](https://github.com/lestrrat-go/jwx/issues/910) -- also see #911). 

---

There could be some improvement in `FromRaw`, but I'm not immediately convinced that it's _necessary_ to achieve what you want. At the moment I believe it's a separate issue from allowing algorithms not present in the JOSE RFCs.

decentralgabe:
Thanks @lestrrat you're right as usual 😄 

[...]

PR #969, opened 2023-08-24
Expose ways to integrate new key types better

lestrrat:
Addresses #912


lestrrat:
I don't like the API that exposes the hooks for the raw keys. I need to rethink it

jwk/jwk.go
```diff
@@ -728,6 +495,20 @@ func CurveForAlgorithm(alg jwa.EllipticCurveAlgorithm) (elliptic.Curve, bool) {
 	return ecutil.CurveForAlgorithm(alg)
 }
 
+// KeySpec is a specification for additional key types
```
lestrrat, line 498: This isn't going to fly. Need to remove

lestrrat:
Notes: need to fix x25519 for > go1.20

lestrrat:
note: if we want to use crypto/ecdh for  go >= 1.20. we're going to have to break back compat for x25519.Public/PrivateKeys, as ecdh.PublicKey/ecdh.PublicKey expects to be treated as pointers, whereas previously we used []byte, which meant that we didn't need to carry pointers.  Aligning this with crypto/ecdh probably means we need to release v3

PR #977, opened 2023-09-08
Sample implementation of jwk.Equal

### LINKED ISSUE #539: Add a way to compare jwk.Key
I need a way to check whether 2 arrays of `jwk.Key` intersect. And to achieve it I iterate over those and compare the keys.

I did not find any way to compare `jwk.Key`s. It would be great to have something like `Equal()` function at that interface. And the implementation would compare essential key fields - the same fields that are used for `Thumbprint()`.

lestrrat:
fixes #539


treeder:
Seems to work for me. 👍

treeder:
@lestrrat could you do a version bump so we can get this function?

PR #990, opened 2023-09-27
Change Get() API

### LINKED ISSUE #826: [v3] RFC: Changing the API for Get()
Because all of jwx objects where allowed to carry both pre-defined and arbitrary fields that appear on the JSON objects, we needed two sets of accessors.

The first group consisted of accessors whose target fields were pre-defined, and their type known (and agreed on by the users) before hand. These field could have nicely named and typed accessors like the following:

```go
token.NotBefore() // time.Time
token.Subject() // string
```

This doesn't work for all the extra fields that the user may specify. Therefore we created the `Get`/`Set` APIs

```go
token.Get(`x-my-very-special-claim`)
```

The problem with this API is that the `Get` API could only return `interface{}`, because we don't know the type of value beforehand. This means that to use this value we still needed one more type conversion:

```go
v, _ := token.Get(`x-my-very-special-claim`)
if s, ok := v.(string); ok {
  // finally got a usable value
}
```

For v3, we'd like to change this signature to

```go
Get(string, interface{}) error
```

This allows you to pass a pointer to a typed variable, if you as the user knows what the value is supposed to be:

```go
// I know, as the person using this particular flavor of token,  that this is a float64
var val float64
if err := token.Get(`x-my-very-special-claim`, &val); err != nil {
   ...
}
// Use val
```

This work much like how `json.Unmarshal` works, and should be relatable to most Go users, and shorten the number of lines that they have to write by a handful of lines.

lestrrat:
Fixes #826


lestrrat:
For whatever reason, we forgot to change jwk.Key's Get() methods :/

PR #994, opened 2023-10-13
add jwk.IsPrivate to check if an asymmetric key is public or private

sding3:
add jwk.IsPrivate to check if an asymmetric key is public or private.


lestrrat:
@sding3 Thank you for your PR.

So, the idea is that for a given asymmetric key, have jwx check if it's private?

```go
if _, err := jwk.ParseKey( jsondata, jwk.WithPrivate(true); err != nil {
   return fmt.Errorf(`key must be private key: %w`, err)
}
```

My initial gut reaction is that it doesn't belong in jwx -- if the RFCs **do** allow symmetric keys, I don't think it's a good idea to allow options that can't be use for all key types in the main entry point of the API.

I'd be more comfortable with something like

```go
package jwk

type isPrivater interface{ // BAD name...
   IsPrivate() bool
} 
func IsPrivateKey(key Key) bool {
  ip, ok := key.(isPrivater)
  if ok {
     return ip.IsPrivate()
  }
  return false
}
```

... and add a `IsPrivate()`   method to the keys, so you can do

```go
key, err := jwk.ParseKey(...)
if key.KeyType() != jwa.OctetSeq {
  if !jwk.IsPrivate(key) {
     return fmt.Errorf(`key must be private (was: %T)`, key)
  }
}
```

Anyway, that's my initial reaction, but please feel free to convince me otherwise.

sding3:
Thank you for the feedback! Agreed that separating this off into its own function makes more sense. I've pushed the update. Only deviation from your suggestion is that I have `IsPrivate` to be able to return an error upon encountering a non-asymmetric key.

jwk/jwk.go
```diff
@@ -744,3 +744,13 @@ func Equal(k1, k2 Key) bool {
 
 	return bytes.Equal(tp1, tp2)
 }
+
```
lestrrat, line 747: We're planning to make it possible for third parties to add their new key types, so I think this interface should be public, allowing new keys to decide for themselves if they want to opt-in to this mechanism.

How about

```go
package jwk

type AsymmetricKey interface {
   IsPrivate() (bool, error)
}
```

jwk/jwk.go
```diff
@@ -744,3 +744,13 @@ func Equal(k1, k2 Key) bool {
...
+// IsPrivate tells whether the supplied key is a private or public key.  An
+// error is raised if the supplied key is not an asymmetric key.
+func IsPrivate(k Key) (bool, error) {
+	ak, ok := k.(asymmetricKey)
```
lestrrat, line 751: I think this check can just be

```go
symmetric, ok := k.(SymmetricKey)
if ok {
   return symmetric.IsPrivate(), nil
}
```

jwk/jwk.go
```diff
@@ -744,3 +744,13 @@ func Equal(k1, k2 Key) bool {
 
 	return bytes.Equal(tp1, tp2)
 }
+
+// IsPrivate tells whether the supplied key is a private or public key.  An
+// error is raised if the supplied key is not an asymmetric key.
+func IsPrivate(k Key) (bool, error) {
```
lestrrat, line 750:
```suggestion
func IsPrivateKey(k Key) (bool, error) {
```

I think that this feels more natural:

```go
key.IsPrivate()
jwk.IsPrivateKey(key)
```

jwk/jwk_internal_test.go
```diff
@@ -191,3 +191,56 @@ func TestIterator(t *testing.T) {
 		})
 	}
 }
+
+func TestIsPrivate(t *testing.T) {
```
lestrrat, line 195: This internal test thing is going to go away, so I'd like to refrain from adding tests in this file. Can you please test through the public API. Yeah, I know it's a bit of a pain, but I'm sure you can find source for all of the key types throughout the test

lestrrat (changes requested):
I added a few comments. If you want, I can fix up the PR for you, but some people don't like that, so... :) just let me know either way.

sding3:
Thanks for the review. I think I've hit all the comments:

- exported jwk.AsymmetricKey interface
- renamed jwk.IsPrivate to jwk.IsPrivateKey
- moved units out of jwk_internal_test.go

lestrrat:
Thanks. LGTM! but please be aware that I won't be making a release immediately afterwards, just to make sure that nothing unintended happens because of this change. Please use specify the appropriate SHA1 from go.mod. And when/if you reach a point where you MUST have a released version and we haven't yet done so, please ping us then.

PR #1011, opened 2023-10-30
Add jwe.WithCEK

lestrrat:


lestrrat:
Although the original poster has not responded, I think this is not a problematic addition, as long as using the static CEK can't be sneaked in by mistake. so I'll merge it

PR #1012, opened 2023-11-08
Custom elliptic curve example

lestrrat:


lestrrat:
I just realized that with the current code at 24daab5c801a77b3bc55295aaba9b7f5c15ee454 we can't convert from jwk.Key to raw key back, especially when we are reusing one of the pre-existing keys. In order to make the round-trip work, we need to be able to inject custom heuristics code in the Raw() method of each object.

lestrrat:
I think this is ready minus some documentation fixes and possible ironing out of some details.

lestrrat:
I don't like the naming. Perhaps JR -> Export, RJ -> Import

lestrrat:
For the record, at this point I think at least 80% of the code works. I'm still missing tests/checks for OKP keys.
I need to revisit and delete dead code, and I need to review the changes and check again if they are user-friendly enough... well, it's not going to be "friendly", because it's complicated, but I hope it's going to be understandable.

lestrrat:
Alright, it's about time. Going to merge to develop/v3

PR #1044, opened 2023-12-19
update all dependencies 12/19/2023

nlacey:
This should fix all of your golang.org/x/crypto security issues.


nlacey:
this should combine PR's 1035-1042

lestrrat:
Thanks. And sorry, this bit isn't as streamlined as it should be: you need to run a bazel //:gazell-update-repos command after upgrading mods. Thankfully, I did put checks for it! :D

I'll do that myself

lestrrat:
merged via #1045

PR #1045, opened 2023-12-19
Merge #1044

lestrrat:


nlacey:
Thanks!  can we get a new release?  thanks for a great tool!

lestrrat:
@nlacey I'm kind of waiting for response on https://github.com/lestrrat-go/jwx/discussions/1034, and either way generally we don't do releases as soon as a fix goes in to wait for the dust to settle, so it will take at least a few days if not more (also, end of year holidays...). If you are in a hurry, for the time being I suggest pegging your builds to a commit hash.

PR #1049, opened 2024-01-07
Add jws.IsVerificationError

### LINKED ISSUE #1048: Expose function to check if `jws.Verify` failed with `verifyError` or not
**Abstract**

If I use `jws.Verify` and it fails, it is hard to know why. If signature is invalid, I would like to return Unauthenticated/Invalid argument error. In all other cases I would like to return just Internal error and log that error.

If I got keys from `jwk.Fetch` and similar, I might be almost 100% sure that issue is because of invalid signature.
But would be nice to know that for sure.

Something similar like https://github.com/lestrrat-go/jwx/blob/develop/v2/jwt/validate.go#L147 or just simple helper

```go
func IsVerifyError(err error) bool {
	var ve verifyError
	return errors.As(err, &ve)
}
```

lestrrat:
fixes #1048

### DIFF
--- a/jws/jws.go
+++ b/jws/jws.go
@@ -477,6 +477,18 @@ func (e *verifyError) As(target interface{}) bool {
 	return false
 }
 
+// IsVerificationError returns true if the error came from the verification part of the
+// jws.Verify function, allowing you to check if the error is a result of actual
+// verification failure.
+//
+// For example, if the error happened while fetching a key
+// from a datasource, that error should return false, whereas a failure to
+// compute the signature for whatever reason would be a verify error
+func IsVerificationError(err error) bool {
+	var ve *verifyError
+	return errors.As(err, &ve)
+}
+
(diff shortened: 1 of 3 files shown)


jws/jws.go
```diff
@@ -477,6 +477,18 @@ func (e *verifyError) As(target interface{}) bool {
 	return false
 }
 
+// IsVerificationError returns true if the error came from the verification part of the
+// jws.Verify function, allowing you to check if the error is a result of actual
+// verification failure.
+//
+// For example, if the error happened while fetching a key
+// from a datasource, that error should return false, whereas a failure to
+// compute the signature for whatever reason would be a verify error
```
arxeiss, line 486: Nitpick: Missing `.` (dot - end of sentence).

arxeiss (approved):

lestrrat:
Thanks for checking. will merge.

PR #1064, opened 2024-01-24
Use io.ReadFull to ensure that generated jwk match the expected keysize

vihangm:
`io.Reader.Read` docs say the following: 

```
If some data is available but not len(p) bytes, Read conventionally returns what is available instead of waiting for more.
```

We need to use `io.ReadFull` or `rand.Read` (which in turn calls `ReadFull`) to ensure that we fill the entire buffer.


lestrrat:
LGTM. Thanks!

Issue #1077, opened 2024-02-19
[v1][v2] Add option to disallow JSON format for JWTs


lestrrat:
jwt.Parse has historically allowed Full JSON format, but RFC7519 does not.
Since we can't change behavior from already released versions, add an option to disallow parsing Full JSON format JWTs.

lestrrat:
After briefly looking at v1, I think it may be harder to change v1 than v2. We may have to punt it for v1

lestrrat:
Decided to punt it for v1. v3 got this through rebase.

PR #1078, opened 2024-02-19
[v2] jws: allow specifing parse format/jwt: disallow JSON

lestrrat:
Solution #1077 for v2

Adds the ability to control which serialization formats are accepted in `jws.Parse` and `jws.Verify`, and further adds `jwt.WithCompactOnly(bool)` to the global settings to disallow parsing of JSON formats.

This makes jwx better adhere to the RFC while keeping backwards compatibility.


lestrrat:
@P3ngu1nW

PR #1127, opened 2024-05-11
Fix building on 32-bit systems

ItalyPaleAle:
Builds on 32-bit systems are failing:

```
 /go/pkg/mod/github.com/lestrrat-go/jwx/v2@v2.0.21/jwt/openid/birthdate.go:69:7: math.MaxInt64 (untyped int constant 9223372036854775807) overflows int
```

This changes how 32-bit systems are detected in a way that doesn't cause the error. Although this does remove 16-bit systems from the `case`, the Go compiler does not support any 16-bit system at present (and it's very unlikely it ever will).

### DIFF
--- a/jwt/openid/birthdate.go
+++ b/jwt/openid/birthdate.go
@@ -60,7 +60,7 @@ func (b *BirthdateClaim) UnmarshalJSON(data []byte) error {
 
 func parseBirthdayInt(s string) int {
 	i, _ := strconv.ParseInt(s, 10, 64)
-	if math.MaxInt == math.MaxInt32 && i > math.MaxInt32 {
+	if i < 0 || (math.MaxInt == math.MaxInt32 && i > math.MaxInt32) {
 		return 0
 	}
 	return int(i)
--- a/jwt/openid/birthdate.go
+++ b/jwt/openid/birthdate.go
@@ -59,11 +59,9 @@ func (b *BirthdateClaim) UnmarshalJSON(data []byte) error {
 }
 
 func parseBirthdayInt(s string) int {
-	var i int64
-	if math.MaxInt == math.MaxInt32 {
-		i, _ = strconv.ParseInt(s, 10, 32)
-	} else {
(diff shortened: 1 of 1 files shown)


ItalyPaleAle:
I'm not sure how to appease CodeQL at this point. It's a false positive but nothing I do seems to make it happy.

lestrrat:
Thanks, but I'm a bit reluctant to merge something that I can't test.
Would you be able to provide a 32-bit environment smoke test, possibly by running a container or some such in GitHub Actions?

ItalyPaleAle:
Sadly GH Actions runs on amd64, and Docker runs 64-bit containers on 64-bit operating systems like our laptops.

Simplest thing you can do to test this, if you're on a Linux box with `qemu-system-arm` and `qemu-user-binfmt` installed (this will be very slow):

```sh
cd examples
GOOS=linux GOARCH=arm GOARM=7 go test
```

The `go test` command above will fail when cross-compiling for armv7 (32-bit), during the build stage, with this error:

```
# github.com/lestrrat-go/jwx/v2/jwt/openid
../jwt/openid/birthdate.go:69:7: math.MaxInt64 (untyped int constant 9223372036854775807) overflows int
```

If you check out this PR, you'll see the build passes and the tests (running using emulation) work.

lestrrat:
@ItalyPaleAle Well, I wanted to make sure I get to test and fix the CodeQL problem.
Also, I'm not talking about you or this PR, but I've had my share of I-trusted-the-PR-without-much-scrutiny-and-it-bit-me-later types of problems, so I'm being cautious here :)

Anyways. I see how you  verified the 32-bit compilation, so I've dug in and isolated the problem down to #1129 . This way there's no CodeQL shenanigans, and I think Go on 32-bit arch can compile it. Does this work for you?

ItalyPaleAle:
Did not take it personally at all, your request was very valid and reasonable :)

#1129 LGTM and I think will fix the problem while making CodeQL happy. I'll close this PR. Thanks!

PR #1142, opened 2024-06-18
Add support for RSA-OAEP-384 and RSA-OAEP-512

Hannes-Kunnen:
This pull request would add support for the `RSA-OAEP-384` and `RSA-OAEP-512` key encryption algorithms. I think this is just a minor change as the base algorithms `RSA-OAEP` and `RSA-OAEP-256` are already implemented. The only difference these new algorithms have is that they use a different hashing algorithm `SHA384` and `SHA512` respectively instead of `SHA1` or `SHA256`.

The hashing algorithms `SHA384`and `SHA512` are already used in different algorithms, so these are not new. The only thing I wasn't sure about it how the following case should be handled:
[permalink](https://github.com/Hannes-Kunnen/jwx/blob/458bea5fa7a67cc251703d04654eb93f8df2bb9f/jwe/internal/keyenc/keyenc.go#L458-L477)
```go
func (e RSAOAEPEncrypt) EncryptKey(cek []byte) (keygen.ByteSource, error) {
	var hash hash.Hash
	switch e.alg {
	case jwa.RSA_OAEP:
		hash = sha1.New()
	case jwa.RSA_OAEP_256:
		hash = sha256.New()
	case jwa.RSA_OAEP_384:
		hash = sha512.New384()
	case jwa.RSA_OAEP_512:
		hash = sha512.New()
	default:
		return nil, fmt.Errorf(`failed to generate key encrypter for RSA-OAEP: RSA_OAEP/RSA_OAEP_256/RSA_OAEP_384/RSA_OAEP_512 required`)
	}
	encrypted, err := rsa.EncryptOAEP(hash, rand.Reader, e.pubkey, cek, []byte{})
	if err != nil {
		return nil, fmt.Errorf(`failed to OAEP encrypt: %w`, err)
	}
	return keygen.ByteKey(encrypted), nil
}
```
In the original code the `'hash'.New()` method was used, so I continued this usage, however when searching where the `SHA384` and `SHA512` algorithms were used already I found an other method which would result in this diff:
```diff
func (e RSAOAEPEncrypt) EncryptKey(cek []byte) (keygen.ByteSource, error) {
-	var hash hash.Hash
+	var hash crypto.Hash
	switch e.alg {
	case jwa.RSA_OAEP:
-		hash = sha1.New()
+		hash = crypto.SHA1
	case jwa.RSA_OAEP_256:
-		hash = sha256.New()
+		hash = crypto.SHA256
	case jwa.RSA_OAEP_384:
-		hash = sha512.New384()
+		hash = crypto.SHA384
	case jwa.RSA_OAEP_512:
-		hash = sha512.New()
+		hash = crypto.SHA512
	default:
		return nil, fmt.Errorf(`failed to generate key encrypter for RSA-OAEP: RSA_OAEP/RSA_OAEP_256/RSA_OAEP_384/RSA_OAEP_512 required`)
	}
-	encrypted, err := rsa.EncryptOAEP(hash, rand.Reader, e.pubkey, cek, []byte{})
+	encrypted, err := rsa.EncryptOAEP(hash.New(), rand.Reader, e.pubkey, cek, []byte{})
	if err != nil {
		return nil, fmt.Errorf(`failed to OAEP encrypt: %w`, err)
	}
	return keygen.ByteKey(encrypted), nil
}
```
I am not sure which one would be preferred (I assumed the method that was already used).

If I missed something or there is a reason these algorithms weren't implemented I would love to hear it.

### DIFF
--- a/internal/jwxtest/jwxtest.go
+++ b/internal/jwxtest/jwxtest.go
@@ -286,7 +286,7 @@ func EncryptJweFile(ctx context.Context, payload []byte, keyalg jwa.KeyEncryptio
 	var keyif interface{}
 
 	switch keyalg {
-	case jwa.RSA1_5, jwa.RSA_OAEP, jwa.RSA_OAEP_256:
+	case jwa.RSA1_5, jwa.RSA_OAEP, jwa.RSA_OAEP_256, jwa.RSA_OAEP_384, jwa.RSA_OAEP_512:
 		var rawkey rsa.PrivateKey
 		if err := key.Raw(&rawkey); err != nil {
 			return "", nil, fmt.Errorf(`failed to obtain raw key: %w`, err)
--- a/jwa/key_encryption_gen.go
+++ b/jwa/key_encryption_gen.go
@@ -30,6 +30,8 @@ const (
 	RSA1_5             KeyEncryptionAlgorithm = "RSA1_5"             // RSA-PKCS1v1.5
 	RSA_OAEP           KeyEncryptionAlgorithm = "RSA-OAEP"           // RSA-OAEP-SHA1
 	RSA_OAEP_256       KeyEncryptionAlgorithm = "RSA-OAEP-256"       // RSA-OAEP-SHA256
+	RSA_OAEP_384       KeyEncryptionAlgorithm = "RSA-OAEP-384"       // RSA-OAEP-SHA384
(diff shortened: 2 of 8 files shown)


lestrrat:
LGTM! I'll merge, but please note that I won't be making a release anytime soon. I usually like to wait and see if anything breaks before making releases.

Thanks!

P.S.

> I am not sure which one would be preferred (I assumed the method that was already used).

It doesn't really matter. They end up calling the same thing. The rest is stylistic, but for this particular part I didn't have a preference :)

PR #1143, opened 2024-06-19
Add IsSymmetric and IsAsymmetric to signature and key encryption algorithms

Hannes-Kunnen:
**Main changes**
Add the `IsSymmetric` method to `SignatureAlgorithm` to mimic the behavior of the existing `IsSymmetric` method of `KeyEncryptionAlgorithm` and add this method to the shared `KeyAlgorithm` interface for ease of use.

Add new `IsAsymmetric` method to both `SignatureAlgorithm` and `KeyEncryptionAlgorithm` to indicate if the algorithm is asymmetric and add this method to the shared `KeyAlgorithm` interface for ease of use. The method is added because `SignatureAlgorithm` has the `NoSignature` option which is neither a symmetric or asymmetric algorithm. `KeyEncryptionAlgorithm` does not have an option like this, but the method is still added to satisfy the `KeyAlgorithm` interface.

Indicate the algorithm type (symmetric or asymmetric) in the element of the file generator instead of a separate map.

**Minor changes**
 - Fix missing packages in import during JWA file generation. At present this isn't actually causing issues because the formatter automagically adds the missing ones, but I think it could become problematic (e.g. in  case of identical package names) (correct me if I'm wrong).
 - Fix small typo
 - Fix `KeyAlgorithm` documentation


lestrrat:
@Hannes-Kunnen Thanks. The partial implementation was probably because `KeyAlgorithm` type is a later addition.

---

I'm not entirely sure about changing the `KeyAlgorithm` interface, precisely because of the presence of NoSignature.
Wouldn't it be more "natural" if we were able to distinguish key algorithms that actually have a notion of symmetry first, and only check for `IsSymmetric()` once you establish that? i.e.

```go
type IsSymmetricer interface {  // Ugh, the naming could be better..
   IsSymmetric() bool
}

// Assume NoSignature does NOT implement IsSymmetric, and the rest does.

func IsSymmetric(alg KeyAlgorithm) bool { // not saying we need jwa.IsSymmetric; this is just for demo purposes
   if alg, ok := alg.(IsSymmetricer); ok {
      return alg.IsSymmetric()
   }
   return false
}
```

In my mind this has clear advantage wrt to maintenance:

1. You cut down the number of functions that need to be implemented in the algorithms by half, b/c you don't need IsAsymmetric
2. You can express algorithms that DO have a notion of symmetry versus those that do not (i.e. NoSignature), and therefore is easier to explain
3. KeyAlgorithm type does not need to change.

WDYT?

Hannes-Kunnen:
I took some time to figure this all out as this issue seems more complex than I initially anticipated. I've done some testing and will be providing my findings below. This is based on what you said and my initial pull request.

If we want to have an interface (e.g. `IsSymmetricer`) that is implemented by every existing `SignatureAlgorithm` and `KeyEncryptionAlgorithm` and NOT by `NoSignature` (a valid `SignatureAlgorithm`) the `SignatureAlgorithm` type has to be split up. To do this I suggest using three types `SignatureAlgorithm`, `SigningAlgorithm` and `NoSignatureAlgorithm`, as described below and in the code block.

- `SignatureAlgorithm`: Represents the various algorithms a payload can use to indicate its signature. (This includes "none")
- `SigningAlgorithm`: Represents the various algorithms that can be used to sign a payload
- `NoSignatureAlgorithm`: Represents the "none" algorithm that indicates no signature is used

```go
// SignatureAlgorithm represents the various algorithms a payload can use to indicate its signature.
// This encapsulates all algorithms that can be used to sign payloads (SigningAlgorithm) and
// NoSignature.
// These are described in https://tools.ietf.org/html/rfc7518#section-3.1.
type SignatureAlgorithm interface {
	String() string
	signatureAlgorithm()
}

// SigningAlgorithm represents the various algorithms that can be used to sign a payload.
// These are described in https://tools.ietf.org/html/rfc7518#section-3.1.
// Keep in mind that NoSignature is not a part of this as no signing can be performed with this value
type SigningAlgorithm string // Implements the signatureAlgorithm() method

// NoSignatureAlgorithm represents the "none" algorithm that indicates no signature is used.
// This is described in https://tools.ietf.org/html/rfc7518#section-3.1.
type NoSignatureAlgorithm string // Implements the signatureAlgorithm() method
```
Note: this change would be **breaking** as users won't be able to cast a `string` to a `SignatureAlgorithm` (`SignatureAlgorithm(string)`) and the `Accept(value interface{}) error` method would not be accessible directly from the `SignatureAlgorithm`. (I still think the change should happen, for reasons I explain below).

Now when it comes to the `KeyAlgorithm` type: I'll try to explain my view of the situation, and maybe you can tell me if and where I go wrong here. This interfaces was originally added to provide a way for users to use to value of the `alg` field of a `jwk.Key` in functions like `jwe.WithKey`. This is because the `alg` field can currently contain either a `SignatureAlgorithm` or a `KeyEncryptionAlgorithm` (or actually any type that implements the `String() string` method. However, (and we go back to the main culprit here) a `jwk.Key` can never have the `NoSignature` value so `KeyAlgorithm` should never accept this value nor any other type that implements the `String() string` method.

This is where `IsSymmetric()` comes in, if the `KeyAlgorithm` interface requires the `IsSymmetric()` method and both `SigningAlgorithm` and `KeyEncryptionAlgorithm` implement it only they will be accepted and functions no longer need to do an extra check if the current `SignatureAlgorithm` is `NoSignature` because the compiler will prevent this (as it should be). It is also possible to add a `keyAlgorithm()` method to the `KeyAlgorithm` interface (and both `SigningAlgorithm` and `KeyEncrpytionAlgorithm` implement it) to prevent other types that implement the `IsSymmetric()` method in the future aren't accepted (this is probably better for maintenance). The use of either the `keyAlgorithm()` or `IsSymmetric()` method in `KeyAlgorithm` seems like a must to me, but if you rather have a separate interface for `IsSymmetric()` I understand. 

The biggest issue with `NoSignature` is that it is not really and algorithms that can be used to sign anything but currently still gets accepted everywhere and then turned down after realising it is `NoSignature`. The only places where `NoSignature` is actually allowed is in the `Algorithm() jwa.SignatureAlgorithm` methods of the `Headers` and `Signer` interfaces of the `jws` package. No function even accepts it, if a user wants to sign with no signature they have to use the `WithInsecureNoSignature()` option which then uses the `noneSigner`.

In conclusion: I think this change would make it way easier for users and for maintenance to know what functions/methods can accept/return `NoSignature` (type `SignatureAlgorithm`) and which only accept/return algorithms that actually sign/verify the payload (`SigningAlgorithm`).

I've implemented this change in a separate branch of my fork. This is just a functional implementation, some files that should be generated aren't and the `genjwa` generator will need a rewrite. I wouldn't mind doing all that, but just not if it turns out that you rather not make this change. Below I linked diffs of two branches, one that only contains code changes that were required to make this work and one that also contains updates to the comments to reflect this change and fixes typos/mistakes I discovered in comments while working through this.

- [Code only diff with develop/v2](https://github.com/lestrrat-go/jwx/compare/develop/v2...Hannes-Kunnen:jwx:v2-rework-none-algorithm-v2-no-comments)
- [Code and comment diff with develop/v2](https://github.com/lestrrat-go/jwx/compare/develop/v2...Hannes-Kunnen:jwx:v2-rework-none-algorithm-v2)

---

Sidenote:
`KeyAlgorithm` can currently (and in my changes) also accept `InvalidKeyAlgorithm`. This means it would also have to implement the `KeyAlgorithm` methods. I just don't quite know why this type exists as its value is always `""` and it is never used separately. The `KeyAlgorithmFrom` function explains that this type exists so this function would not have to return an error and so the result can be used immediately in another function. However, if we simply remove this type and return `nil` instead (as `KeyAlgorithm` is an interface) the same result is achieved. I'm just adding this here as this could be part of the same change (it would also be **breaking** (see code block)).
```go
algorithm := jwa.KeyAlgorithmFrom("")
// Old way
if algorithm != jwa.InvalidKeyAlgorithm("") {
	// Do something
}
// New way
if algorithm != nil {
	// Do something
}
```

lestrrat:
First of all, thanks for taking the time to review this in such depth. I'm pretty impressed you took the time to dig deep into this codebase's guts :)

So while in principle I agree that there could be better organization/modeling of these algorithms,  in general I can't accept breaking changes. There are users other than you, so breaking API changes need to be planned for major version updates. In order for them to go-in immediately, the status quo should have a real negative impact to the users. 

You can _add_ APIs/types pretty safely (aliasing is also acceptable). You can't remove APIs/types. And you can't expect me to accept any code where users need to change their code without a compelling reason.

I'd like to reiterate that I'm in general in line with your _intent_. However, if you want to break the API, please consider doing so for v3 (develop/v3). Or if you feel strongly about including your changes in v2, you're going to have to be more creative and/or make concessions with the fact that you can't expect users to change existing code.

---

That said, here are comments to specific points.

> However, (and we go back to the main culprit here) a jwk.Key can never have the NoSignature value so KeyAlgorithm

That's true. And this confusion came up because the implementation of KeyAlgorithm came up much later in the development. 

I don't remember it anymore, but I could imagine myself thinking that you needed to handle "none" after only reading the RFC only a couple of times -- much later I realized that it needed to be prohibited in general, but allowed only if the user forcibly told us to do so.

> The biggest issue with NoSignature is that it is not really and algorithms that can be used to sign anything but currently still gets accepted everywhere and then turned down after realising it is NoSignature

Ditto. Fair enough. I'm okay with the idea of separating NoSignature from the rest in v2, **but _ONLY_ if old code still works**. That is to say, A new definition with new type can be added. However, the constant NoSignature must stay, with very strong discouragement to use it anywhere. Also, where users were previously allowed to (e.g. jws.Parse), it needs to be accepted, and needs to end up with a valid jws.Message object (you probably need to manually implant the old NoSignature type instead of whatever new thing you add). Also you need to allow users to access/modify its value when they want to tweak JWS headers.

> KeyAlgorithm can currently (and in my changes) also accept InvalidKeyAlgorithm

I'm sure I meant it to be some sort of a zero value for key algorithm, but in hindsight it probably could have gone away. Unlike changes you mentioned including type name changes, since I do agree that not many people will use it even on accident, we can probably warn ahead of time that this will  go away at some point, and remove it in a few months time. So doc patches discouraging its use and making sure that no part of the current codebase use it would be good, but it needs to stay for now.

Hannes-Kunnen:
I probably should have made my changed against the develop/v3 branch, that's why I tried to make it clear the change would be breaking.

> I'm okay with the idea of separating NoSignature from the rest in v2, but ONLY if old code still works.

I don't see a way to do this without introducing a breaking change, if we just add a new type for `NoSignature` type casting will still be required. (I'm fairly new to go, so maybe there is a way that I haven't figured out).

> So while in principle I agree that there could be better organization/modeling of these algorithms, in general I can't accept breaking changes.
> I'd like to reiterate that I'm in general in line with your _intent_.

Does this mean I can take the same approach when creating a possible pull request for the develop/v3 branch?

> Also you need to allow users to access/modify its value when they want to tweak JWS headers.

I don't really understand what you mean with this, is there some place I forgot to allow this? Or is this related to just separating `NoSignature` in v2?

> Unlike changes you mentioned including type name changes, since I do agree that not many people will use it even on accident, we can probably warn ahead of time that this will go away at some point, and remove it in a few months time.

It still is a breaking change and I would assume some people will check the result of the `KeyAlgorithmFrom`. When I was playing around with changing this I discovered tests would fail because they use a generic string as input to the `KeyAlgorithmFrom` function when setting the the `alg` field and later expect it to contain that value (which means the headers will have an invalid `alg` field). This leads me to believe that there are probably scenarios where users SHOULD check this to not create invalid headers without realizing it (typos). (Changing it to `nil` would create headers without the `alg` field, which might make it more obvious that something went wrong). The way these users should prepare for this change is instead of checking if the the type is `InvalidKeyAlgorithm` they should check if it's either `SignatureAlgorithm`or `KeyEncryptionAlgorithm`. At that point they could just use their individual accept methods instead of `KeyAlgorithmFrom`.

Changing this only in v3 would remove the potential of breaking existing setups. It depends on what you prefer and on when an official v3 release is expected.

---

For now it seems these changes could be made:
**v2**

- Add the `IsSymmetric()` method to `SignatureAlgorithm`(in this or a new pull request) and warn users of the following caveats:
    - When checking if a `SignatureAlgorithm` is asymmetric an extra check should be made to check if the algorithm is NOT `NoSignature`
        ```go
          if alg != jwa.NoSignature && !alg.IsSymmetric() {
            // alg is asymmetric
          }
        ```
    - Manually registered algorithms (`Register...Algorithm()`) will always return false (so these should be checked separately), this applies to both `SignatureAlgorithm` and `KeyEncrpytionAlgorithm` (registering these can change in v3 to allow the user to indicate if it is symmetric or not)
- Add the `IsSymmetric()` method to the `KeyAlgorithm` interface (with the same warnings) (in the same pull request)
  - This will no longer allow types other than `SignatureAlgorithm` or `KeyEncryptionAlgorithm` to be seen as a `KeyAlgorithm` (even though `keyAlgorithm()` should be the real indicator). This change should not really be breaking as all functions currently fail if they receive other types.
- Add the `keyAlgorithm()` method to the `KeyAlgorithm` interface and make `SignatureAlgorithm`and `KeyEncryptionAlgorithm` implement it to indicate they are key algorithms (in a new or the same pull request).
  - This change should not really be breaking as all functions currently fail if they receive other types.
  - Maybe add disclaimer that `NoSignature` shouldn't really be a `KeyAlgorithm` and that his will change in v3
- Add new pull request to fix mistake/typos I discovered in comments while researching this issue

**v3**

- Create a pull request to differentiate `SigningAlgorithm` and `NoSignatureAlgorithm` that implement the `SignatureAlgorithm` interface (like I did based on the develop/v2 branch)
    - Also update registering `KeyEncryptionAlgorithm`and the new `SigningAlgorithm` to allow the user to indicate if it is a symmetric algorithm. We would not allow the user to register new `NoSignatureAlgorithm` values (unless you think this can be useful).
- Create pull request to remove the `InvalidKeyAlgorithm` type

Let me know if you would change anything in these proposed changes and how you think they should be separated (pull request wise).

lestrrat:
tl;dr:

* Since you say you can't do it in v2 without breaking API compatibility, there's no possibility of me taking this PR. As such I won't comment further on your proposed changes to v2.
* For v3, the bar is much lower, and sure, you can break the API. Your proposed changes looks okay, but I still would have to see it before I can commit anything.
* For v3, you would have to include more docs/examples to show the changes and migration path to existing users.

P.S. if you are making changes to v3, I think I'm not 100% up-to-date with porting changes from v2 to v3, so watch out.

---

**Main reply:**

> Does this mean I can take the same approach when creating a possible pull request for the develop/v3 branch?

Ah, probably? I'm sorry if I sound non-committal, but it's hard to give you a clear answer without evaluating the proposed changes as a whole.

Before I go on, I'd like to put your proposed changes in perspective: Yes, the flaws that you mention are valid. But at the same time, they are not critical in the sense that they are not something that prevents somebody to do what they want. For the **user**, it's not super important what type NoSignature has. It just needs to work.  And your proposed changes touch a relatively large number of places for a non-critical issue. The combination of the criticalness of the problem and the impact of the changes are one of the few reason why I'm not straight out accepting your changes.

Now, getting back to the question of would I accept the same change for v3:
For v2, I could see how it would impact current users, so I could say no immediately with just your code.
But for v3, the context is going to be slightly different, because we would need to show the existing v2 users what their migration path is going to be. You probably need to include docs on what changed and why, and show how users need to change their code from v2. The docs might end up being rather small, so this might be not such a bit problem, but it's important that you have thought it through the eyes of the users. And only then would I be able to evaluate the PR as a whole.

Sort of off topic, but I'd like to stress the value I'm looking for. For changes that require action on the users' part, what matters is not the correctness of the modeling: it's the right balance of correctness and ease of use for the users. In this case, ease of use would include how easily existing users can understand the new model and migrate to the new version, including examples and explanation of concepts. If for some reason a change for correctness hinders the ease of use, then we need to find a concession. Because I think your change is non-critical, it's extra important that you be extra nice to the users in this regard.

> ... when an official v3 release is expected.

TBH I don't expect v3 to be released anytime soon, as there is no significant motivation for people to upgrade to it. 
Currently the changes in v3 are only cleanups (including your changes) on existing modeling/design, so there's no compelling reason for people (or the maintainers) to immediately move to v3. Reasons could include limitation in the API, or perhaps something to do with the underlying Go runtime/version, but until such a motivation is found, it's probably going to remain an unofficial, experimental version.

(You are always welcome to suggest a reason - in a separate thread - to expedite the release, if there is one!)

---

**Minor stuff:**

>> Also you need to allow users to access/modify its value when they want to tweak JWS headers.
>
>I don't really understand what you mean with this, is there some place I forgot to allow this? Or is this related to just separating NoSignature in v2?

No, TBH I didn't check. I just stated that to stress that you can't just fudge with the types if it breaks existing setter/getters, if any exist. 

> Manually registered algorithms (Register...Algorithm()) will always return false

If you're doing it for v3, this probably should be remedied. Users should be able to specify it.

Hannes-Kunnen:
> Since you say you can't do it in v2 without breaking API compatibility, there's no possibility of me taking this PR. As such I won't comment further on your proposed changes to v2.

What I meant was that I can't separate `NoSignature` from the rest in v2, the changes I proposed after for v2 were related to this pull request and should not be breaking.

I just added a new commit to make part of these changes, I think you were okay with these. What I wanted to know was if you are okay with adding `IsSymmetric()` and `keyAlgorithm()` to the `KeyAlgorithm` interface as you seem hesitant about changing it. I've added that commit to a separate branch, you can see the diff with this pull request [here](https://github.com/Hannes-Kunnen/jwx/compare/v2-genjwa-is-symmetric-asymmetric...Hannes-Kunnen:jwx:v2-genjwa-is-symmetric-asymmetric-v2). I don't think these changes are breaking. The only thing it changes is that users can no longer pass any type that implements the `String()` method to functions that accept `KeyAlgorithm` (doing so would always result in an error while casting to `SignatureAlgorithm` or `KeyEncryptionAlgorithm` but would now be blocked by the compiler).

lestrrat:
Looks *almost* good to me !

One thing I think you can do is to actually remove this restriction:

> https://github.com/lestrrat-go/jwx/pull/1143/files#diff-c2439f931b7a6e1c34aa6296f4785add4154071eeb4bf050b363045f51dd359fR138



```go
// RegisterKeyEncryptionAlgorithm registers a new KeyEncryptionAlgorithm so that the jwx can properly handle the new value.
// Duplicates will silently be ignored
//
// You may pass extra options to tweak the behavior of this function. For example, you can pass `WithSymmetricAlgorithm(true)` to let the library know that it's a symmetric algorithm.
func RegisterKeyEncryptionAlgorithm(v KeyEncryptionAlgorithm, options ...RegisterKeyEncryptionAlgorithmOption) {
	var symmetric bool
		switch option.Ident() {
		case identSymmetricAlgorithm{}:
			symmetric = option.Value()
		}
	}
	muKeyEncryptionAlgorithms.Lock()
	defer muKeyEncryptionAlgorithms.Unlock()
	if _, ok := allKeyEncryptionAlgorithms[v]; !ok {
		allKeyEncryptionAlgorithms[v] = struct{}{}
		if symmetric {
			symmetricEncryptionAlgorithms[v] = struct{}{}
		}
		rebuildKeyEncryptionAlgorithm()
	}
}

func (v KeyEncryptionAlgorithm) IsSymmetric() bool {
       _, ok := symmetricAlgorithms[v]
      return ok
}

// lots of details omitted.
```

Variadic options effectively allows us to change the API without having the user needing to change their code (unless of course they are doing `var MyKeyEncrptionAlgorithm func(jwa.KeyEncryptionAlgorithm)` for some god awful reason). Alternatively we could add a new function named `RegsiterKeyEncryptionAlgorithmWithOptions()`, whose name is meh but really doesn't break anybody's code. I guess the latter is safer.

WDYT?

Hannes-Kunnen:
I think that because invalid options can be provided (not all algorithms can accept the `WithSymmetricAlgorithm`option), an error has to be returned, so the use of a separate function is required to prevent breaking changes. I've implemented this change and also added extra tests for the existing register functions and the new with options variant.

Let me know if you like the change and whether I can also add the change to the `KeyAlgorithm` interface.

lestrrat:
> I think that because invalid options can be provided (not all algorithms can accept the WithSymmetricAlgorithmoption), an error has to be returned

I think you can get away by only implementing those that make sense to be implemented. For example `RegisterEllipticCurveAlgorithmWithOptions`, doesn't need to exist.

I think you can also state that you document the new function(s) are stop gap functions that will eventually be deprecated and merged with the main `RegisterXXXX` functions.

Hannes-Kunnen:
I've implemented this change, but am wondering how new options would be handled in the future? If no error can be returned will all options that can not be accepted by one of the functions just have to be ignored?

lestrrat:
@Hannes-Kunnen Sorry, I was away for vacation, and am now trying to catch up with life in general. I should come back to this soon enough, but just not yet. I'm letting you know so that you are not just left hanging.

lestrrat:
@Hannes-Kunnen I added some minor tweaks to your branch in https://github.com/lestrrat-go/jwx/tree/Hannes-Kunnen-v2-genjwa-is-symmetric-asymmetric. If you're okay, I'll merge your changes from my branch.


> I've implemented this change, but am wondering how new options would be handled in the future? If no error can be returned will all options that can not be accepted by one of the functions just have to be ignored?

I think you have a point, but at the same time I don't think we can plan for something that we don't know what it's going to be used for. In my tweak I've added that this function should be considered experimental, and not a stable API, which I believe should be a warning enough for users who want to use it.

Hannes-Kunnen:
I'm okay with your changes, but would like to reiterate my request to change the `KeyAlgorithm` interface.
I've added my proposed changes in the spoilers below (not the generated files). If you don't want these changes that's okay, but it seems you haven't reacted to them yet. The reason for adding the `keyAlgorithm()` method is to limit the accepted algorithms to `SignatureAlgorithm` and `KeyEncryptionAlgorithm` using the compiler instead of an error response (this will prevent users from having to run the code and realise they are using the wrong type). The `IsSymmetric()` method is just an ease of use addition as all key algorithms implement it, but can be left out.

<details>
<summary>jwa/jwa.go</summary>

```diff
  type KeyAlgorithm interface {
+     keyAlgorithm()
      String() string
+     // IsSymmetric returns true if the algorithm is a symmetric type.
+     // Keep in mind that the NoSignature algorithm and the InvalidKeyAlgorithm type will always
+     // return false as these do not indicate symmetry.
+     IsSymmetric() bool
  }

  // InvalidKeyAlgorithm represents an algorithm that the library is not aware of.
  type InvalidKeyAlgorithm string

+ func (InvalidKeyAlgorithm) keyAlgorithm() {}
+
  func (s InvalidKeyAlgorithm) String() string {
	  return string(s)
  }

+ func (InvalidKeyAlgorithm) IsSymmetric() bool {
+     return false
+ }
+
  func (InvalidKeyAlgorithm) Accept(_ interface{}) error {
      return fmt.Errorf(`jwa.InvalidKeyAlgorithm does not support Accept() method calls`)
  }
```
</details>

<details>
<summary>tools/cmd/genjwa/main.go (changes generated files jwa/key_encryption_gen.go and jwa/signature_gen.go)</summary>

```diff
  func _main() error {
      typs := []typ{
          ...
          {
              name:      `SignatureAlgorithm`,
              comment:   `SignatureAlgorithm represents the various signature algorithms as described in https://tools.ietf.org/html/rfc7518#section-3.1`,
              filename:  `signature_gen.go`,
              symmetric: true,
+             keyAlg:    true,
              elements: []element{
                  ...
              },
          },
          {
              name:      `KeyEncryptionAlgorithm`,
              comment:   `KeyEncryptionAlgorithm represents the various encryption algorithms as described in https://tools.ietf.org/html/rfc7518#section-4.1`,
              filename:  `key_encryption_gen.go`,
              symmetric: true, 
+             keyAlg:    true,
              elements: []element{
              ...
              },
          },
      }
      ...
  }

  type typ struct {
      name      string
      comment   string
      filename  string
      elements  []element
      symmetric bool
+     keyAlg    bool
  }

...

  func (t typ) Generate() error {
      ...
+    
+     if t.keyAlg {
+         o.LL("// keyAlgorithm is used to qualify as a KeyAlgorithm.")
+         o.L("func (%s) keyAlgorithm() {}", t.name)
+     }
      ...
  }
```
</details>

lestrrat:
@Hannes-Kunnen I thought you retracted those. Do you want to continue on this thread? That just means this PR isn't going to be merged until this is all resolved.

---

Off the top of my head:

re: Adding `keyAlgorithm()`: Makes sense if jwx only allows algorithm types which are defined within the library itself. I suppose that's OK if we do NOT allow `RegisterXXXX` functions to accept user-defined algorithms ... Is there any chance users might need more than a string representation of an algorithm?.... probably not, so probably okay, though I'm still a bit reluctant to gratuitously change API at this moment. I'd be going "YES!" if I were getting repeated questions about what can be passed, but because I'm not getting any real life examples of this, I'm inclined to wait for next v2 minor release or v3.

re: Adding `IsSymmetric()` to `KeyAlgorithm`: no. This being a public API (unlike `keyAlgorithm()` above), this is not happening until next v2 minor release or v3.

re: Adding `IsSymmetric()` to `InvalidKeyAlgorithm`: Seems harmless, but [I'm not convinced that it's strictly necessary](https://github.com/lestrrat-go/jwx/pull/1143#issuecomment-2179584461). I'd still prefer this approach:

```go
type IsSymmetricer interface {  // Ugh, the naming could be better..
   IsSymmetric() bool
}

func IsSymmetric(alg KeyAlgorithm) bool { // not saying we need jwa.IsSymmetric; this is just for demo purposes
   if alg, ok := alg.(IsSymmetricer); ok {
      return alg.IsSymmetric()
   }
   return false
}
```

Hannes-Kunnen:
I don't mind dropping them both. But I'm not sure why adding `keyAlgorithm()` would prevent any of the current actions from happening. The `RegisterXXXX` functions currently accept only their specific type e.g. `RegisterSignatureAlgorithm` only accepts a `SignatureAlgorithm` and should be used like this (the variable is not necessary):
```go
newAlgorithm := jwa.SignatureAlgorithm("new-algorithm")
jwa.RegisterSignatureAlgorithm(newAlgorithm)
```
When a JWK is unmarshalled it uses the `jwa.KeyAlgorithmFrom(string)` function which will detect this is a `SignatureAlgorithm` (as we registered it) and thus a `KeyAlgorithm` and return it. So it can be used anywhere it should be and the user can register  `SignatureAlgorithm` and/or `KeyEncryptionAlgorithm`, which are the two types the `KeyAlgorithm` interface was designed for. Unless I'm missing something?

Anyway, if you rather not implement this that's okay, it's not something that has a significant impact like you said but can be useful for people new to the library.

lestrrat:
As I said, I'm in agreement that it's probably OK, but I'm saying it does not sound like something that I feel compelled to include right at this moment, because it adds no immediate value (and nobody is asking for it). 

And because of that, I'm saying it should probably be planned for next v2 minor release or v3.

Hannes-Kunnen:
Okay, then let's drop it.
Just wondering (for future reference), how would something be planned for a v2 minor release?

lestrrat:
@Hannes-Kunnen FYI https://github.com/lestrrat-go/jwx/tree/develop/v2.2. I have no concrete plans for this branch, but throwing it out there so you have something to work with

lestrrat:
Since this PR is going to be closed soon (b/c I'm merging), might want to create a new issue to discuss it.

lestrrat:
closing because I merged #1151. Let us continue discussion elsewhere if need be.

PR #1170, opened 2024-08-12
Fix link to openid

flokli:
This is in the `openid` subdirectory inside the `jwt` directory, but this file is already inside the latter.

### DIFF
--- a/jwt/README.md
+++ b/jwt/README.md
@@ -8,7 +8,7 @@ Package jwt implements JSON Web Tokens as described in [RFC7519](https://tools.i
 * Conversion to and from JSON
 * Generate signed tokens
 * Verify signed tokens
-* Extra support for OpenID tokens via [github.com/lestrrat-go/jwx/jwt/openid](./jwt/openid)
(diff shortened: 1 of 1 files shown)


lestrrat:
thanks!

lestrrat:
@flokli oh, sorry, you made this against "main", but we don't use main (I should delete it, but it exists for historical reasons). You were probably should have created it against develop/v2

I'll revert this PR

flokli:
That's the default branch GitHub suggested. Maybe it's something in the GitHub settings?

lestrrat:
unfortunately, nope :)
image

flokli:
Hm, then I might have ended up reading on GitHub in the main branch for some reason. It's a bit confusing.

Assuming all commits on that branch are in another branch too, to prevent them being GCed, I'd prefer for the main branch to be removed, so this won't happen to others ending up there the same way I did.

lestrrat:
It's been a while, but I believe the v0 code lives in that branch. Realistically we should be able to remove it, but I think it'd be more prudent to do cleanups in the go.mod files to mark deprecations, etc first -- which I'm not really up to doing ATM (yes, admittedly out of sheer laziness)

PR #1216, opened 2024-10-18
update comments to include iat

cscochris:
`WithAcceptableSkew()` and `WithClock()` both apply to `iat` as well as `nbf` and `exp` - update comments to reflect that.

### DIFF
--- a/jwt/options.yaml
+++ b/jwt/options.yaml
@@ -45,7 +45,7 @@ options:
     interface: ValidateOption
     argument_type: time.Duration
     comment: |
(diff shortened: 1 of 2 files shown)


lestrrat:
Ah, I didn't immediately realize the nuance in your original discussion post: I thought you were only talking about "if `iat` needs to be checked", but I now I see it was more of a spec(comment) vs implementation thing. Thank you.

I'll do one round of sanity check before merging to see if this is indeed a comment problem and not an implementation problem.

lestrrat:
Right, yeah, these were just comments predating the actual code. Thanks for the fixes.

PR #1308, opened 2025-02-25
jws/jwe: split token into fixed number of parts

drakkan:
this avoid to use eccessive memory when processing maliciously crafted tokens with a large number of '.' characters.

See also:

https://github.com/golang/go/issues/71490
https://github.com/go-jose/go-jose/security/advisories/GHSA-c6gw-w398-hv78

### DIFF
--- a/jwe/jwe.go
+++ b/jwe/jwe.go
@@ -886,10 +886,11 @@ func parseJSON(buf []byte, storeProtectedHeaders bool) (*Message, error) {
 }
 
 func parseCompact(buf []byte, storeProtectedHeaders bool) (*Message, error) {
-	parts := bytes.Split(buf, []byte{'.'})
-	if len(parts) != 5 {
-		return nil, fmt.Errorf(`compact JWE format must have five parts (%d)`, len(parts))
+	// Five parts is four separators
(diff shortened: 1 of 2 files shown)


lestrrat:
@drakkan If this is done for safety, I wonder if using `bytes.SplitN` using `n + 1` as the number of expected segments would be better?

For example

```go
input := []byte("one.two.three")
expectedSegments := 3
split := bytes.SplitN(input, []byte{'.'}, expectedSegments+1)

if len(split) != expectedSegments {
   return ...
}
```

lestrrat:
@drakkan Oh, I think I get it. You want to avoid creating that extra slice of bytes, which _could_ contain potentially very large amounts of garbage. I guess in that case Count would be better.

drakkan:
> @drakkan Oh, I think I get it. You want to avoid creating that extra slice of bytes, which _could_ contain potentially very large amounts of garbage. I guess in that case Count would be better.

yes, thank you. 

The Go project and also jose reported this as a security issue. I think it's a bit excessive but if possible a new version with the fix would help (I'm fine using an unreleased tag until this will happen). Thanks in advance

PR #1309, opened 2025-02-26
jws/jwe: split token into fixed number of parts (#1308)

lestrrat:
this avoid to use eccessive memory when processing maliciously crafted tokens with a large number of '.' characters

refs #1308 , for v2 line


drakkan:
an improvement for this fix is under discussion [here](https://go-review.googlesource.com/c/oauth2/+/655455)

PR #1322, opened 2025-03-20
jwk: fix missing increment in set.Keys to correctly fill the return slice

johejo:
fix https://go.dev/play/p/u0rjrarzeTM


lestrrat:
I'll merge after adding slight tweaks in a separate PR

lestrrat:
merged via #1323

PR #1347, opened 2025-04-10
Add the claim name to error messages produced by ClaimValueIs()

jmalloc:
This PR adds changes the behaviour of `ClaimValueIs` to include the claim name in the error message produced when the claim key is present, but the value does not match the required value. The message produced when the claim is absent already included the name.

The existing test cases cover this scenario by checking that the error is non-nil, but as there were no existing assertions about the message _content_, I chose to leave these as is.

### DIFF
--- a/docs/01-jwt.md
+++ b/docs/01-jwt.md
@@ -2,35 +2,34 @@
 
 In this document we describe how to work with JWT using `github.com/lestrrat-go/jwx/v3/jwt`
 
-* [Terminology](#terminology)
-  * [Verification](#verification)
-  * [Validation](#validation)
-* [Parsing](#parsing)
-  * [Parse a JWT](#parse-a-jwt)
-  * [Parse a JWT from file](#parse-a-jwt-from-file)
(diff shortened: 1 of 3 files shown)


lestrrat:
@jmalloc I think it's just the error messages not matching, but the tests are falling. Can you please make sure you run the entire test suite? Thanks.

jmalloc:
My bad, I didn't realise there were multiple Go modules in the repo. I've successfully executed the tests using `make test` now instead of simply running `go test`.

lestrrat:
@jmalloc Yeah, sorry, it's a bit confusing. I didn't want to include modules that are only used in examples/ dir in go.mod

lestrrat:
@jmalloc merged, but I'm not sure if I want to make a release with just this change, so I hope you're okay with commit-hash-pinning until I do.

jmalloc:
> so I hope you're okay with commit-hash-pinning until I do.

Understandable, and no worries. Thanks!

PR #1360, opened 2025-04-18
jws: improve performance for SplitCompact/SplitCompactString

drakkan:
name            old time/op    new time/op    delta
SplitCompat-16    76.5ns ± 5%    21.1ns ± 1%   -72.40%  (p=0.000 n=10+10)

name            old alloc/op   new alloc/op   delta
SplitCompat-16      272B ± 0%        0B       -100.00%  (p=0.000 n=10+10)

name            old allocs/op  new allocs/op  delta
SplitCompat-16      2.00 ± 0%      0.00       -100.00%  (p=0.000 n=10+10)

### DIFF
--- a/jws/jws.go
+++ b/jws/jws.go
@@ -632,17 +632,26 @@ func parseJSON(data []byte) (result *Message, err error) {
 	return &m, nil
 }
 
+const tokenDelim = "."
+
 // SplitCompact splits a JWT and returns its three parts
 // separately: protected headers, payload and signature.
 //
 // On error, returns a jws.ParseError.
 func SplitCompact(src []byte) ([]byte, []byte, []byte, error) {
-	// Three parts is two separators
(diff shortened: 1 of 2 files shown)


lestrrat:
@drakkan thanks!

drakkan:
Thanks! Could this be applied to v2 as well, please?

EDIT:  JWE has a similar splitting mechanism. I don't use JWE myself, but if you'd like, I can try submitting a PR for it as well

lestrrat:
@drakkan I can't cherry-pick this change to v2 due to some problems (*), can you please create a PR against v2? To be clear, they are minor problems that I can deal myself, but I thought you might want to make the commits so that your name will show up in the records, since you were one who came up with this.

Also, I would happily accept it if you submitted a PR for JWE.

(*) biggest one is that there's no jws.ParseError in v2

PR #1362, opened 2025-04-20
jwe: use strings.Cut in parseCompact

drakkan:


lestrrat:
@drakkan thanks!

Note to self: I think I may want to introduce jwe.ParseError here

PR #1380, opened 2025-06-01
Add `b64: false` header to detached example

f3ndot:
When using JWS to sign a detached webhook payload I intend to transmit in the clear, I found the receiving service written in NodeJS (using the very popular `jose` library) assumed the payload was base64-encoded. It demanded the `b64: false` header be sent in order to treat it as plaintext.

While RFC 7797 states this header is OPTIONAL, in order to maximize compatibility with other software and libraries, this example should endeavour to use it. I'm going on the assumption that detached payloads are not likely manually base64-encoded

### DIFF
--- a/examples/jws_sign_detached_payload_example_test.go
+++ b/examples/jws_sign_detached_payload_example_test.go
@@ -24,7 +24,7 @@ func Example_jws_sign_detached_payload() {
 	hdrs := jws.NewHeaders()
 	hdrs.Set("b64", false)
 	hdrs.Set("crit", []string{"b64"})
-	
+
 	serialized, err := jws.Sign(nil, jws.WithKey(jwa.HS256(), key, jws.WithProtectedHeaders(hdrs)), jws.WithDetachedPayload([]byte(payload)))
 	if err != nil {
 		fmt.Printf("failed to sign payload: %s\n", err)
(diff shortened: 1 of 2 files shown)


examples/jws_sign_detached_payload_example_test.go
```diff
@@ -17,7 +17,15 @@ func Example_jws_sign_detached_payload() {
...
 	}
 
-	serialized, err := jws.Sign(nil, jws.WithKey(jwa.HS256(), key), jws.WithDetachedPayload([]byte(payload)))
+	// If you plan to transmit your payload without base64-encoding (RFC 7797),
+	// it's best to set `b64: false` so libraries can act accordingly (e.g. the
+	// popular NodeJS `jose` library requires it set to false in order to verify
+	// a plaintext payload.
+	hdrs := jws.NewHeaders()
+	hdrs.Set("b64", false)
+	hdrs.Set("crit", "b64")
```
f3ndot: Also, I'm fairly new to both Go and JWx stuff, but I think this `crit` header is ignored?

lestrrat:
@f3ndot Thanks, but I think you haven't ran the examples: you need to actually change directories to `examples`, and then run `go test`.

The line after `// OUTPUT:` need to be `// eyJhbGciOiJIUzI1NiIsImI2NCI6ZmFsc2UsImNyaXQiOlsiYjY0Il19..lnRw_MSpQjARa5LWqPcu8Qls9p3wYGrC6tz4-nr0rkA`

f3ndot:
@lestrrat my apologies, I was using the GitHub editor for the quick PR. I've updated the tests now.

lestrrat:
merged via #1383

PR #1381, opened 2025-06-02
Use jwk interfaces for each key type rather than concrete types

### LINKED ISSUE #1379: jwk.Import for embedded jwk.Keys
I've seen implementations in the wild where jwk.Key is embedded in another struct:

```go
type Foo struct {
   jwk.Key
   ... fields ...
}
```

And this happily passes as `jwk.Key`, obviously.

But since we're comparing actual types in jwk.Import, this doesn't work (`*Foo` != `*jwk.ecdsaPrivateKey`, etc).

I think, instead of checking for actual types, check if the key _acts_ like the concrete type that we're expecting.

lestrrat:
fixes #1379


jwk/convert.go
```diff
@@ -351,6 +351,15 @@ func bytesToKey(src interface{}) (Key, error) {
...
 }
 
+type Embedded struct {
+	Key
+}
+type DoubleEmbedded struct {
+	Embedded
+}
+
+var _ Key = &DoubleEmbedded{}
```
lestrrat: Thus was for debugging. need to delete

jwk/jwk.go
```diff
@@ -631,3 +631,42 @@ type keyWithD interface {
...
+	if rv.NumField() == 0 {
+		return nil, fmt.Errorf(`invalid key value type %T (2): %w`, keyif, ContinueError())
+	}
+	// Iterate through the fields of the struct to find the first field that
+	// implements the Key interface
+	for i := range rv.NumField() {
+		field := rv.Field(i)
+		if field.CanInterface() {
+			if k, ok := field.Interface().(Key); ok {
+				return k, nil
```
lestrrat: I think we can recursively call extractEmbeddedKey here

PR #1396, opened 2025-06-10
all: use `any` type alias instead of `interface{}`

zchee:
Go introduced `any` in 1.18.
The lestrrat-go/jwt's go.mod is `go1.23`, and now gopls and another linters detect it as a "diagnostic", so I replaced everything, including the output of each gen tool.


lestrrat:
@zchee Sorry, I just did a major revert. I'll close this for now, because I expect a lot of changes underneath. I'll make the changes you requested when I can.

PR #1453, opened 2025-09-05
[jwk] Fix possible connection leak in Fetch (#1452)

Warh40k:
* close http.Response.Body right after checking err


lestrrat:
merged via #1458

PR #1465, opened 2025-10-03
Change go.mod version requirements to go 1.24.0 and introduce toolchain directive

henrymcconville:
Currently the go version is 1.24.4 which sets a minimum version of 1.24.4 where 1.24.0 should be sufficient. This downgrades the go version in all go.mod files and introduces the `toolchain` directive to maintain a consistent build version.

### DIFF
--- a/bench/comparison/go.mod
+++ b/bench/comparison/go.mod
@@ -1,6 +1,8 @@
 module github.com/lestrrat-go/jwx/bench/comparison
 
-go 1.24.4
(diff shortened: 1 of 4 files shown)


lestrrat:
Ah yes. `go mod tidy` or whatever just keeps changing it for me, so I kind of resigned to the idea that the tools are going to do what the tools are going to do, but yeah, if it's causing real problems, by all means...

PR #1485, opened 2025-11-02
Fix null support

### LINKED ISSUE #1484: when parsing, string registered claims should not accept null
**Describe the bug**

When `jwt.Parse` parses a token which include one of the registered claims (except aud), it accepts value as `null` which is per JSON specification a type on its own (not a string)

Please attach the output of `go version`: `go version go1.24.9 linux/arm64`

```
package main

import (
        "log/slog"

        "[github.com/lestrrat-go/jwx/v3/jwa](http://github.com/lestrrat-go/jwx/v3/jwa)"
        "[github.com/lestrrat-go/jwx/v3/jwk](http://github.com/lestrrat-go/jwx/v3/jwk)"
        "[github.com/lestrrat-go/jwx/v3/jwt](http://github.com/lestrrat-go/jwx/v3/jwt)"
)

const secret = "abracadabra"

func main() {
        key, err := jwk.Import([]byte(secret))
        if err != nil {
                slog.Error("error creating jwk", slog.Any("err", err))
                return
        }
        for k, v := range map[string]string{
                "no_sub":   "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.e30.O9ZhNCv3pn22mroAcowAq5Z0OiaMEHV43nJdrk76Qj4",
                "null_sub": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOm51bGx9.Wab9LGF3D0_ogJ39wRk-Ga6SeFZCKAe3cU4s6xDL21o",
                "null_aud": "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhdWQiOm51bGx9.ehhQNg6N1MGJ3ceEvMPMfHl2nZOQR_aCvAVMXXx8BOM",
        } {
                token, err := jwt.ParseString(v, jwt.WithKey(jwa.HS256(), key))
                if err != nil {
                        slog.Error("error parsing token", slog.String("name", k), slog.Any("err", err))
                        continue
                }
                sub, has_sub := token.Subject()
                slog.Info("parsed token", slog.String("name", k), slog.Any("payload", token), slog.String("sub", sub), slog.Bool("has_sub", has_sub))
        }
}
```

I expect null_sub to be rejected. Other string registered claims are also affected (jti and iss).

> tokens in example above have been generated using jwt.io

**Additional context**
 
Will submit a PR

cboitel:
Closes #1484

### DIFF
--- a/internal/json/json.go
+++ b/internal/json/json.go
@@ -46,11 +46,16 @@ func AssignNextBytesToken(dst *[]byte, dec *Decoder) error {
 }
 
 func ReadNextStringToken(dec *Decoder) (string, error) {
-	var val string
-	if err := dec.Decode(&val); err != nil {
-		return "", fmt.Errorf(`error reading next value: %w`, err)
+	token, err := dec.Token()
+	if err != nil {
+		return "", fmt.Errorf(`error reading next token: %w`, err)
 	}
-	return val, nil
+	switch v := token.(type) {
+	case string:
+		return v, nil
+	default:
+		return "", fmt.Errorf("invalid type: %T", token)
+	}
 }
 
 func AssignNextStringToken(dst **string, dec *Decoder) error {
--- a/jwt/jwt_test.go
+++ b/jwt/jwt_test.go
@@ -1640,3 +1640,76 @@ func TestGH1482(t *testing.T) {
 	require.NoError(t, err, `jwt.Parse should succeed`)
 	require.NotEmpty(t, markerValue, "context value 'marker' should be present")
 }
+
+func TestNullValues(t *testing.T) {
+	key, err := jwk.Import([]byte("abracadabra"))
+	require.NoError(t, err, "jwk key import should succeed")
+
+	testCases := []struct {
+		name   string
+		input  string
+		errStr string
+	}{
+		{
+			name:   `"iss"=null`,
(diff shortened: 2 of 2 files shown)


lestrrat:
We've already gone over this. This cannot be the default behavior. 

image

cboitel:
Sorry I and you may have been confused.

Just to be clear, date processing was clearly an issue regarding default behaviour since pedantic is disabled by default. But regarding null, this is a different subject. I haven't modified your tests, just added some new. It proves that default behaviour isn't broken by this fix.

Would a fix using an option `WithNullValuesDisabled(true/false)` be accepted (default value would be false) ?

lestrrat:
Your change breaks backwards compatibility. 

Try adding the following test to the vanilla v3.0.12. This passes.

```go
// in jwt/jwt_test.go
func TestNilValue(t *testing.T) {
	const payload = `{"iss":null}`

	key := []byte("secret")
	jws.Sign([]byte(payload), jws.WithKey(jwa.HS256(), key))

	_, err := jwt.Parse([]byte(payload), jwt.WithVerify(false))
	require.NoError(t, err, `jwt.Parse should succeed`)
}
```

Now, apply your changes to `json.ReadNextStringToken()`, and run the same test.

image

---

Also, [I have a work in progress solution](https://github.com/lestrrat-go/jwx/tree/pedantic)
This is the third time I mention this ([first time here](https://github.com/lestrrat-go/jwx/security/advisories/GHSA-xx5p-7j8f-mqv9#advisory-comment-142850), [second time here](https://github.com/lestrrat-go/jwx/issues/1486#issuecomment-3479116182))

PR #1799, opened 2026-04-14
copy recipientKey in KeyDecryptAESGCMKW to avoid caller aliasing (JWE-002)

lestrrat:
Fix JWE-002. `KeyDecryptAESGCMKW` built its GCM ciphertext with
`ciphertext := recipientKey[:]; append(ciphertext, tag...)`, which mutates the
backing array of `recipientKey` whenever there's spare capacity. Since
`recipientKey` is the parsed Message's owned encrypted-key slice, this
corrupts message state on multi-recipient JWEs or any re-decrypt. Allocate
a fresh buffer instead.

Regression test in `jwe/jwebb/decrypt_test.go` constructs a backing array
with sentinel tail bytes and asserts they're untouched; it fails on the
prior code and passes after the fix.

v3 counterpart: #TBD

### DIFF
--- a/jwe/jwebb/decrypt_test.go
+++ b/jwe/jwebb/decrypt_test.go
@@ -69,6 +69,33 @@ func TestKeyDecryptAESGCMKW(t *testing.T) {
 	require.Equal(t, cek, decrypted)
 }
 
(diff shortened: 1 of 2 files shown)


lestrrat:
Cross-link: v3 counterpart is #1800.

PR #1980, opened 2026-04-18
jwe: enforce WithMaxParseInputSize in Decrypt

lestrrat:
## Summary
- `jwe.Decrypt` now rejects `buf` exceeding `WithMaxParseInputSize` (default 10MB) before any JSON/compact parsing or crypto, matching `jwe.Parse*`
- `WithMaxParseInputSize` is accepted on `DecryptOption` for per-call override via new `GlobalParseDecryptOption` interface
- Closes the DoS gap where `json.Unmarshal` ran on attacker-sized input before `maxRecipients` was checked

## Test plan
- [x] `go test -race ./jwe/...`
- [x] `go vet ./...`
- [x] `golangci-lint run ./jwe/...`

### DIFF
--- a/jwe/jwe.go
+++ b/jwe/jwe.go
@@ -324,6 +324,7 @@ type decryptContext struct {
 	dst                     *Message
 	maxRecipients           int
 	maxDecompressBufferSize int64
+	maxParseInputSize       int64
 	maxPBES2Count           int
 	minPBES2Count           int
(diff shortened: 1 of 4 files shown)


lestrrat:
Superseded by #1984 — capping raw parse input bytes is the caller's responsibility; see the stance doc at docs/13-input-size.md in that PR.

PR #2063, opened 2026-04-27
jwa: unify SignatureAlgorithm/KeyEncryption/ContentEncryption into one registry

lestrrat:
Backport of #2062.

`KeyAlgorithmFrom("X")` used to walk three independent registries — `SignatureAlgorithm` → `KeyEncryptionAlgorithm` → `ContentEncryptionAlgorithm` — and return the first hit. There was no cross-registry uniqueness, so an extension could register the same name in two kinds and `KeyAlgorithmFrom` would silently flip resolution with import order.

Funnels all three KeyAlgorithm-implementing kinds through a single shared `algRegistry` (hand-written in `jwa/jwa.go`). Per-kind `Register*` / `Lookup*` / `Unregister*` / `<Kind>s()` become thin wrappers around five private helpers (`registerAlgorithm`, `markBuiltin`, `unregisterAlgorithm`, `lookupAlgorithm`, `listAlgorithmsByKind`) emitted by the v3 generator (`tools/cmd/genjwa`).

Differences from the v4 PR:
- v3's `Register*` returns nothing, so cross-kind collision **panics** with a structured message at the conflicting init() call site (the only way to surface the conflict in the v3 signature).
- v3's pre-existing per-kind built-in replacement was a silent no-op; that's preserved (v3 documents this as frozen behavior; v4 escalates it to an error).

Same-kind idempotent re-registration is preserved.

Bundles JWA-002: `KeyAlgorithmFrom` typed arms now reject zero-value typed algorithms (`String() == ""`) with `ErrInvalidKeyAlgorithm`.

Out of scope:
- `KeyType`, `EllipticCurveAlgorithm`, `CompressionAlgorithm` — they don't dispatch through `KeyAlgorithmFrom` and don't have JWA-001's collision shape.

Tests added at `jwa/cross_kind_test.go`: cross-kind collision (3 directions, asserted via `require.PanicsWithValue`), unambiguous lookup, zero-value rejection (3 typed kinds + `EmptySignatureAlgorithm()`), idempotent re-register.

### DIFF
--- a/jwa/content_encryption_gen.go
+++ b/jwa/content_encryption_gen.go
@@ -5,17 +5,9 @@ package jwa
 import (
 	"encoding/json"
 	"fmt"
-	"sort"
-	"sync"
 
 	"github.com/lestrrat-go/jwx/v3/internal/tokens"
 )
-
-var muAllContentEncryptionAlgorithm sync.RWMutex
-var allContentEncryptionAlgorithm = map[string]ContentEncryptionAlgorithm{}
-var muListContentEncryptionAlgorithm sync.RWMutex
-var listContentEncryptionAlgorithm []ContentEncryptionAlgorithm
-var builtinContentEncryptionAlgorithm = map[string]struct{}{}
 
 func init() {
 	// builtin values for ContentEncryptionAlgorithm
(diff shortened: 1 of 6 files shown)


lestrrat:
Closing — the panic on cross-kind collision is a behavior change v3 shouldn't take. Will reconsider the right v3 treatment separately.

PR #2210, opened 2026-06-18
document slow-path alg handling in jws verify

lestrrat:
- document that jws.Verify (slow path) intentionally does not enforce header alg == verifier alg; verifier alg comes from key/provider
- correct misleading VerifyCompactFast godoc line about where the algorithm comes from
- add maintainer comment on verifyContext.tryKey explaining the omission is intentional, not an algorithm-confusion bug
- doc-only; no behavior or signature change


lestrrat:
Superseded: instead of documenting the slow-path alg leniency, we will enforce protected-header alg == verifier alg in tryKey by default (with a WithSkipAlgorithmMatch opt-out). New PR to follow.

PR #2220, opened 2026-06-18
document Chain.Get returns aliased read-only slice

lestrrat:
- document that cert.Chain.Get returns a slice aliasing internal storage that callers must treat as read-only
- intentionally no defensive copy: alias never crosses a trust boundary and corruption is only caller-self-inflicted; comment-only, no behavior change

### DIFF
--- a/cert/chain.go
+++ b/cert/chain.go
@@ -94,6 +94,17 @@ func (cc *Chain) UnmarshalJSON(data []byte) error {
 // Get returns the n-th ASN.1 DER + base64 encoded certificate
 // stored. `false` will be returned in the second argument if
 // the corresponding index is out of range.
+//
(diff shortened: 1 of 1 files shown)


lestrrat:
Gate note: across four independent codex review passes this comment-only PR received only **unactionable** verdicts — one out-of-diff (findings about files this PR does not touch) and three empty `NOT SATISFIED` with no findings body. Per the review-gauntlet bailout rule (codex returning repeated unactionable verdicts), the gate was completed by a maintainer-conducted adversarial self-review of the one-paragraph diff: the comment is accurate (`Get` returns `cc.certificates[index]`, which genuinely aliases internal storage; the read-only/trust-boundary reasoning holds; no behavior change). Labeled `gauntlet-accepted` on that basis.

PR #2275, opened 2026-07-25
jwe: serialize the JSON "aad" member as BASE64URL(JWE AAD)

sueun-dev:
`Message.MarshalJSON` writes the JSON `"aad"` member as `base64std(protected_header + "." + BASE64URL(aad))` instead of `BASE64URL(JWE AAD)`. RFC 7516 §7.2.1 defines the member as the external Additional Authenticated Data on its own, base64url-encoded; the protected-header concatenation belongs only to the AEAD input (§5.1/§5.2 step 14), which the encrypt and decrypt paths already build separately with `concatAAD`.

Two things are wrong in the emitted member:

- the protected-header segment and `.` are prepended, and
- `aad` is passed to `marshalField` as a `[]byte`, so `json.Marshal` base64-encodes it a second time, in the padded std alphabet — sibling members `ciphertext`/`iv`/`tag` pass a string from `base64.EncodeToString`, so only `aad` is affected.

A JWE JSON message that carries an external `aad` (as produced by any RFC-compliant library) therefore does not survive `jwe.Parse` -> `json.Marshal` -> `jwe.Parse`: the aad decodes to the wrong value with no error, because the lenient decoder accepts the mangled string. `jwe.Encrypt` has no aad option, so the trigger is the parse-then-reserialize / forward path, not encrypt.

This is the semantic sibling of the `MarshalJSON` aad block touched in #1794 — that change fixed slice aliasing in the same spot but left the value itself wrong.

What changed:

- `Message.MarshalJSON` encodes `aad` as a base64url string, like the other members, with no protected-header prefix.
- added `TestJWEJSONAADRoundTrip`, pinning the RFC 7516 §7.2.1 member value and the `Parse` -> `Marshal` -> `Parse` round trip.

Checked (`GOEXPERIMENT=jsonv2`, go 1.26.4):

- `go test ./jwe/...` — pass; `TestJWEJSONAADRoundTrip` fails before the change (`aad` member is `ZXlKbGJt...==`), passes after.
- `go test ./...` — pass.
- `go vet ./jwe/`, `gofmt -l` — clean.

### DIFF
--- a/jwe/message.go
+++ b/jwe/message.go
@@ -247,12 +247,14 @@ func (m *Message) MarshalJSON() ([]byte, error) {
 	}
 
 	if aad := m.AuthenticatedData(); len(aad) > 0 {
-		aad = base64.Encode(aad)
-		if encodedProtectedHeaders != nil {
-			aad = concatAAD(encodedProtectedHeaders, aad)
-		}
-
-		v, err := marshalField(aad)
+		// RFC 7516 §7.2.1: the "aad" member is BASE64URL(JWE AAD) — the
+		// external Additional Authenticated Data on its own. The protected
+		// header is prepended to the AAD only when building the AEAD input
+		// (concatAAD, in the encrypt/decrypt paths), never in the serialized
+		// member. Encode to a base64url string like the ciphertext/iv/tag
+		// members above so marshalField does not base64-encode the raw bytes
+		// a second time (which also used the padded std alphabet).
+		v, err := marshalField(base64.EncodeToString(aad))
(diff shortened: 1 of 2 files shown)


lestrrat:
Thanks for the fix, but let me think about it, as I'm sure this an area I haven't given much thought on. I may want to bundle a few more fixes

For the record, #1794 may have introduced the double encoding, but this has been a long-standing bug, as it was introduced in #212, 6 yrs ago. Goes to show not too many people use JSON serialization :D

PR #2297, opened 2026-08-17
Fix build under Go 1.27 (json/v2.SkipFunc removal)

moznion:
Go 1.27 removed encoding/json/v2.SkipFunc and handed its role to errors.ErrUnsupported: json/v2 now decides whether a custom marshal or unmarshal function declined a value with errors.Is(err, errors.ErrUnsupported) instead of the identity comparison err == SkipFunc.

ref:

- https://github.com/golang/go/issues/74324
- https://go-review.googlesource.com/c/go/+/745041
- https://pkg.go.dev/encoding/json/v2@go1.27rc3

internal/json referenced jsonv2.SkipFunc directly, so v4 fails to compile for anyone on Go 1.27:

    internal/json/registry.go:73:18: undefined: jsonv2.SkipFunc

No single expression satisfies both toolchains. Go 1.26 compares the sentinel by identity, so it has to be SkipFunc verbatim, and that identifier no longer exists in 1.27. Bind it to errSkipFunc behind a Go-version build tag instead, keeping go.mod at go 1.26.0 so existing users are unaffected. The shim can be deleted once the module requires Go 1.27.

---

Should I include the CI configuration for 1.27 in this pull request? It’s still in the RC stage, so it might be a bit premature.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -199,7 +199,9 @@ Use `github.com/stretchr/testify/require` for assertions (not `assert`).
 
 ## Build Tags
 
-No build tags in v4. Optional features (signature algorithms, backend replacements) are provided as extension modules under [`github.com/jwx-go`](https://github.com/jwx-go). See [Extension Modules](docs/10-extensions.md) for the full list.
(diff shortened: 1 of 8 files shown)


lestrrat:
@moznion thanks. I will have to take a look at this tomorrow, as I'm out today. But the changes look OK to me.

lestrrat:
merged via #2304

moznion:
Thank you for your work!

